PR #16, opened 2020-06-19
Minimum Node.js version

Aarowaim:
The `fs-extra` package enforces a minimum version for Node.js. Earlier versions will crash the build script due to syntax errors in `fs-extra`; the syntax of `try-catch` has changed in a breaking way.


Eranziel:
Thanks for helping update the docs!

Issue #3, opened 2020-06-05
[FEATURE] Item-type Sheets


Eranziel:
Create sheet templates and logic for the Item-types:
- [x] Skill Triggers
- [x] Talents
- [x] Core Bonuses
- [x] Pilot Equipment
- [x] Mech Frames
- [x] Mech Systems
- [x] Mech Weapons
- [x] Licenses

Grygon:
Tags should also be added as an item-type, given the quantity available, quantity of data involved, and likely desire to implement custom ones.

I'll take a look at whipping that up.

Eranziel:
I was hoping to keep tags more abstract as a simple array property of relevant items, since they are just a name and sometimes a value. Plan was to use `lancer-data` to grab descriptions when needed. Part of the motivation there is that I have no intention of replacing or replicating all of Comp/Con's functionality, and part is it seems like overkill to have Items owned by Items owned by Actors etc... ;-)

That said, I haven't done much work yet with the items other than generating the compendiums. Is there a case I'm missing that would be better handled by making Items for tags?

Grygon:
The big advantage would be the standardization of tags--as an Item, tags could be standardized to a compendium. which could then be pulled from to build weapons/systems/etc.

I guess you're right that it may be adding a bit more complexity than is needed though, both in the chain of ownership and longer-term maintenance. Especially since I didn't know about `lancer-data`, that's super nice to have and will simplify things!

I'll take a look at throwing together some templates for the items above then.

Grygon:
Taking a first pass of each type in [my items branch](https://github.com/Grygon/foundryvtt-lancer/tree/items), then pull them in and then _maybe_ do a second pass to make them pretty, although that's really not an area I'm strong in.

Grygon:
All sheets listed here are complete, but not sure if we want to close this yet since we're still missing the 3 NPC item sheets

Eranziel:
I'll close it and open a new issue for the NPC items.

PR #35, opened 2020-07-10
Better tags & array control. Pilot item sheets, barebones license sheet, and some frame sheet work.

Grygon:
Looks like I did it right this time...


Eranziel:
Closes #3.

PR #102, opened 2020-11-06
Grygon/expanded macros

### LINKED ISSUE #10: [FEATURE] Expanded Macros
Implement macros which print effects text in a fancy chat block. Similar to Roll20 character sheet behaviour.
1. Pilots:
- [x] Skill triggers
- [x] Talents
- [x] Mech systems
- [x] Core bonuses
- [x] Core system active
- [x] Core system passive
- [x] Pilot gear
- [ ] ~~Mech traits (are these necessary?)~~ Removing from scope, unless feature is requested by users.
2. NPCs:
- [x] Weapons
- [x] Tech Attacks
- [x] Reactions
- [x] Traits
- [x] Systems

Grygon:
Extensive changes and expansions to the macro framework.

Closes #10

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ dist/**/*.js.map
 dist/**/*.json
 dist/**/*.css
 dist/fonts/mdi/**
+debug.log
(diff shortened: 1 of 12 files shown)


Grygon:
Was an accidental merge into dev, but from my testing it looks like only one chunk of code went wrong! Do still need to figure out the roll button question...

PR #118, opened 2020-11-29
Reworked overkill tag

HiddenWings25:
Overkill tag now uses a combination of exploding dice and keep highest to achieve repeated rerolls. Overkill reroll button was removed, but exploded dice gives an accurate count of heat in the attack card.


Eranziel:
This has been merged into the dev branch - thanks! The PR will close automatically when that gets merged into master for the next release.

PR #121, opened 2020-12-06
[RDY] Area of Effect Templates

BoltsJ:
A port of the spell templates of the DnD5e system to lancer. There are some changes to the invocation to account for the differences in data and snapping rules are changed.

To create a template, invoke the static method `game.lancer.canvas.WeaponRangeTemplate.fromRange({options})`. The `options` object requires a `type` key that a string and a `val` key that is the size of the template. The options object is of the same schema as the objects contained in the range array on weapons.

A compendium of (hopefully) all the templates found if the core book and Long Rim is included for users, as well as a reference of how to use the class.

Burst templates have a separate snapping function that snaps them to the center of hovered tokens and on-the-fly adjusts the size to conform to the size of the hovered token.

Further work would involve adding these to sheets and/or attack cards.


Grygon (changes requested):
The file "dist/packs/aoe-templates.db" - should be in the src folder instead (src/packs/aoe-templates.db) otherwise it could get wiped out!

BoltsJ:
The gulpfile must be all messed up, because the clean task does nothing for me, but I moved it all the same.

Grygon (approved):

Issue #125, opened 2020-12-18
[FEATURE] Add support for maestro item sounds


kreisl:
Currently the lancer system does not support automatic sounds from the maestro module when rolling weapon attacks.
It would be good to add support for weapon and system-based roles using maestro.
I have provided a basic implementation of this feature for weapon sounds which works for maestro 0.7.2 and lancer 0.1.19.
I am not sure how systems and tech attacks could be implemented in this system as it is designed for items.
link to maestro: https://github.com/death-save/maestro

PR #126, opened 2020-12-18
Added compatibility of weapon rolls with the maestro item-roll feature

kreisl:
This is related to Issue #125.
I am not a ts developer, so i am not sure if it is properly implemented.
This is mainly to show where the changes need to be made for the sounds to work.


Eranziel:
Thanks for the PR to demonstrate, that's very useful!

Please change the PR to target the `dev` branch instead of `master`, as we only update master when making a release.

Issue #120, opened 2020-12-02
[FEATURE] Provide default images for actors


Grygon:
**Describe the solution you'd like**
To make actor (Pilot & NPC) creation easier, we should be able to set the default image for a Mech, which would automatically set that mech's image instead of the generic actor icon. Ideally we also could do this for NPC classes, although that isn't included in LCP/CC data.

Part 1: 

- Take the URL from LCP ~> Frame -> image_url and associate it with the frame. If a token image isn't manually set in the prototype, use this image for any PC.
- Add a field on the mech to maintain this

Part 2: 

- Set up the ability to manually set the image on NPC classes just like the mech option
- Find some default images for all base NPC classes that 1. don't have licensing issues and 2. look cool & would be accepted by most people, then set those images to default for those classes.

**Describe alternatives you've considered**
Manual workflows are fine but this would be very cool

**Additional context**
Totally fine to do this myself but wanted to put this here to make sure it stays on our radar when I inevitably forget

**Discord contact info**
Grygon#9138

Grygon:
Note: This seems to run into CORS issues when trying to access the standard images in Long Rim. We'd need to re-host, which is a whole bag of worms. Putting this project off for a while.

ScyllPoesis:
Big fan of retrograde. Would be nice to put a little link to their stuff in the welcome page and maybe a credits link under the mech avatar on the sheet if we get permission to include their free tier assets.

Grygon:
Addressing in #312

PR #130, opened 2020-12-30
Set actor and token image using Comp/Con or default

flukowski:
This should implement the PC side of https://github.com/Eranziel/foundryvtt-lancer/issues/120.

It will also pull in the frame image if that is set.


src/module/actor/util.ts
```diff
@@ -168,19 +169,34 @@ export async function update_pilot(pilot: LancerActor, cc_pilot: mm.Pilot): Prom
...
+
     }
-    
+
+    // Setup token
+    let mech_image = am.CloudImage || get_mech_art(am.Frame.ID) || pilot.data.img;
+    // FIXME: Setting type as any to allow editing the token attributes
+    let ptoken = duplicate(pilot.data.token) as any;
+
+    pad.img = mech_image;
```
BoltsJ, line 184: The mech image shouldn't be stored over the pilot portrait.

src/module/actor/util.ts
```diff
@@ -168,19 +169,34 @@ export async function update_pilot(pilot: LancerActor, cc_pilot: mm.Pilot): Prom
...
+    ptoken.name = cc_pilot.Callsign;
+    ptoken.img = mech_image;
```
BoltsJ, line 186: This is unconditionally overwriting the token image. This should only happen if the token is already using a default image.

BoltsJ (commented):
The pull request should be against the `dev` branch, not the main branch

Grygon (changes requested):
Marking this as request changes per Bolts' comment

PR #135, opened 2021-03-10
Add custom themes

Lrizika:
Adds a themes directory with custom themes
\- Includes two themes from COMP/CON plus several others
Each theme file defines a number of css variables
\- If desired, custom themes can also add arbitrary css
\- Currently, the only such css they include is scrollbar coloring
lancer.css attempts to resolve these variables
\- GMS Red is hardcoded in as a fallback
updateTheme (defined in settings.ts) loads in the appropriate theme file
\- This is called once in the ready hook
\- It's also called whenever a client updates their theme

Adds a per-client theme setting to settings.ts

Plus a few minor CSS tweaks beyond what's needed for themes


Grygon:
Sorry for the delays here, but with the work getting 0.2 ready haven't been able to look at this.

On that note, would you mind changing this to target `dev-0.2` and handling the necessary merge conflicts from that? A lot has changed but hopefully won't be too bad to tweak it to work.

Grygon (changes requested):
Actually making this a formal request for changes since I'm cleaning up our PRs--should target `dev-0.2`

PR #136, opened 2021-03-11
Remove LANCER.sys_name, reduce hardcoded paths

Lrizika:
Replaced LANCER.sys_name with game.system.id
- This means it takes the system id from system.json

Changed most of the hardcoded /systems/lancer paths to use the system id
- A few of these are harder to replace, so for now I've left them as-is
- On the whole, as long as you have *a* lancer system installed at that
 directory it still works.
- It would be nice to remove hardcoded paths entirely, though

With these changes, it mostly works to run the lancer system under a
different system id and directory, or alongside another instance of the
system. The hardcoded paths are the primary remaining obstacle to this
working entirely, as there will be some cross-contamination (and if you
don't have a lancer system at /systems/lancer, it will just break).


whitespine:
@Eranziel any reason this was never merged?

Eranziel:
Just that it targets `dev`, but we were working on `dev-1.0` already. It probably needs to be rewritten for `dev-1.0`.

Eranziel:
FYI, implemented this on `eran-migration-staging`. There's still some paths that we can't convert unless we generate them in the init hook or similar, since `game.system` doesn't exist until then.

Issue #166, opened 2021-04-30
[BUG] NPC Tech Actions Don't Roll


Grygon:
**Describe the bug**
![image](https://user-images.githubusercontent.com/647846/116716205-1576d100-a99d-11eb-973f-a140d56d0467.png)

**Discord contact info**
Flint

ScyllPoesis:
Fixed in #167, as well as added proper automation and structure. I also added some more improvements to collapses and refactored the player mech sheet Tech Attack over to use the same macro handler.

Grygon:
Closing - Fixed in 0.9.3

PR #167, opened 2021-04-30
tech attack macro fixes for #166

ScyllPoesis:


ScyllPoesis:
Also now including fixes for Flex mounts displaying misleading information in certain configurations, and an implementation of COMP/CON's hex controls for limited system uses.

ScyllPoesis:
Also also including NPC sheet tracking for limited and recharge feats, and limited automation (button to roll recharge for all NPC feats, auto-roll when NPC starts their turn).

PR #209, opened 2021-06-04
MVP for vault imports

SohumB:
image

image

caveats all around, but this should _work_, be reasonably usable, and be able to serve as a good base for the future


src/aws-exports.ts
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable */
+// WARNING: DO NOT EDIT. This file is automatically generated by AWS Amplify. It will be overwritten.
```
Eranziel, line 2: Is this actually an auto-generated file, or copy-pasted into the project from an auto-generated original?

src/module/compcon.ts
```diff
@@ -0,0 +1,33 @@
+import { PackedPilotData } from 'machine-mind';
+import Auth from "@aws-amplify/auth";
+import Storage from "@aws-amplify/storage";
+
+// we don't cache anything other than id/name; we're going to fetch all other data on user input
+// but we don't want the pilot actor window to wait for network calls
+// this does mean that the GM needs to refresh foundry to clear this cache if they add pilots
+// (they could also re-login, we initiate a cache refresh there)
```
Eranziel, line 8: Maybe we can add a refresh button next to the vault pilot selector on the pilot sheet. That would be a nice-to-have, can be done later.

src/module/settings.ts
```diff
@@ -29,6 +30,15 @@ export const registerSettings = function () {
     type: Object,
   });
 
+  game.settings.registerMenu(LANCER.sys_name, "compconLogin", {
+    name: "Comp/Con Login",
+    label: "Log in to Comp/Con",
+    hint: "Log in to Comp/Con to automatically load any pilots and mechs you have access to",
+    icon: "fas fa-bars",
+    type: CompconLoginForm,
+    restricted: true
```
Eranziel, line 39: Since the data storage is all local, I think all users should have access to this. Probably need to make the setting either user-level or client-level, testing now.

src/aws-exports.ts
SohumB: It's copy-pasted from the autogenerated original, yes. I'll defer to your judgement about whether it makes sense for the comment to exist or not.

Eranziel: It's fine, just curious. 😉

PR #232, opened 2021-06-22
[Incomplete] Partial fix for Actor item list removal.

purringInsanity:
Actor Item entries get destroyed after calling GenControlContext with data-action="null".

TODO: 
 -Items still left over from Mech Weapon Mounts being reset. 
 -Dropping Items Directly into Expected slots, leave ghost items with broken keys.

### DIFF
--- a/src/module/actor/mech-sheet.ts
+++ b/src/module/actor/mech-sheet.ts
@@ -173,7 +173,7 @@ export class LancerMechSheet extends LancerActorSheet<EntryType.MECH> {
       this._event_handler("reset-wep", evt);
     });
 
-    html.find(".reset-ll-weapon-mounts-button").on("click", async (evt: JQuery.ClickEvent) => {
+    html.find(".reset-all-weapon-mounts-button").on("click", async (evt: JQuery.ClickEvent) => {
       this._event_handler("reset-all-weapon-mounts", evt);
     });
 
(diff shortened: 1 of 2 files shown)


whitespine:
Haven't read this too thoroughly (so forgive me if I'm stating the obvious), but for gen-controls, using data-action="delete" (as opposed to "null") will delete the item outright. 

That won't fix the drag-replacing case but is generally useful. For the drag replace you might want to specifically tweak the ref_dropping handler to allow hooking of pre/post drop, but that might be overkill.

purringInsanity:
Swapping to "delete" had been my original attempt, but I found when using the same weapon multiple times on a mech, delete would remove the weapon from the item list, but would remain on the Mech Actor sheet until all copies of the weapon was removed. I found storing an item, then nulling its slot, followed by destroying the item, did not have this issue.

Grygon:
To confirm--"incomplete" here means it's still a draft right? Or are you calling it ready to go?

whitespine:
For posterity's sake, to explain the observed behavior:

When deleting an item, the loadout slot wasn't actually being cleared. It was simply expected that removing the item would cause the reference to become invalid.

Unfortunately, references are semi-resilient against such changes, in that they will attempt to fallback to any item matching the LID of the original reference. Thus, deleting one pistol would simply make the loadout grab another one.

whitespine:
Before it is merged though, I do want to make a change. Rather than altering the behavior of "null" or "delete", I'd prefer to just add a new "null_delete" option that does the behavior of both. Not every deletion is specifically safe to null out, and vice versa.

purringInsanity:
Closing this PR as discussion has shown this to be working as intended.

PR #239, opened 2021-06-28
Localization pass on actors and chat cards

Aarowaim:
Chat cards and actors are now mostly localized. Some handlebar helpers will need localization logic, and items haven't received a full localization pass. Merry christmas! 🎁

### DIFF
--- a/src/templates/actor/pilot.hbs
+++ b/src/templates/actor/pilot.hbs
@@ -14,12 +14,12 @@
         </div>
 
         <div class="details flexcol">
-          <span>{{"pilot-sheet.header.name"}}</span>
+          <span>{{localize "lancer.pilot-sheet.header.name"}}</span>
           <span class="medium">{{actor.name}}</span>
         </div>
 
         <div class="details flexcol">
-          <span>{{"pilot-sheet.header.bg"}}</span>
+          <span>{{localize "lancer.pilot-sheet.header.bg"}}</span>
           <span class="medium">{{{mm.Background}}}</span>
         </div>
       </div>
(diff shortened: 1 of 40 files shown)


src/templates/chat/overkill-reroll.hbs
```diff
@@ -1,7 +1,6 @@
 <div class="card clipped-bot" style="margin: 0px;">
   <div class="lancer-header ">
-    OVERKILL
-    <button class="overkill-reroll"><i class="fas fa-redo"></i> +d6</button>
+    {{{localize "lancer.chat-card.overkill.reroll"}}}
```
Aarowaim, line 3: This part may require some testing or conversion into a handlebar partial template.

src/templates/actor/pilot.hbs
```diff
@@ -4,23 +4,23 @@
...
 
         <div class="details flexcol">
-          <span>CALLSIGN:</span>
+          <span>{{localize "lancer.pilot-sheet.header.cs"}}</span>
           <span class="pilot-callsign">{{mm.Callsign}}</span>
         </div>
 
         <div class="details flexcol">
-          <span>NAME:</span>
+          <span>{{"pilot-sheet.header.name"}}</span>
```
Grygon, line 17: Interesting that this works... does seem to, but I'm curious how.

AaronLip: First instinct is that it's a typo. Booting up the sheet, there are two empty spans resulting from it; one for name and one for background.

It would need to be `{{localize "pilot-sheet.header.name"}}`. I'd love if handlebars had a strict mode to help catch issues like this.

Aarowaim: First instinct is that it's a typo. Booting up the sheet, there are two empty spans resulting from it; one for name and one for background.

It would need to be `{{localize "lancer.pilot-sheet.header.name"}}`. I'd love if handlebars had a strict mode to help catch issues like this.

It should now work as intended.

Issue #256, opened 2021-07-16
[BUG] HP and heat constrained when changing from the token HUD


BoltsJ:
**Describe the bug**
When changing HP and heat values from the token hud, the value is constrained by the defined min and max. This is a regression from 0.1.20 where HP could go into the negative.

**To Reproduce**
1. Create a mech and assign a frame.
2. Deploy the token
3. Set the HP to 0 from the HUD
4. Subtract 1 HP from the HUD by typing `-1` in the box
5. Observe that the value goes to 0.

**Expected behavior**
HP should be able to go below 0 and heat should be able to go above the cap.

**Screenshots**
![image](https://user-images.githubusercontent.com/182554/125886530-d1cfeb7c-ba70-4312-b7a7-ac1669c22cbd.png)

**Platform (please complete the following information):**
 - OS: Pop_OS 21.04
 - Browser firefox
 - Foundry version 0.8.8
 - Lancer system version 0.9.5

**Additional context**
This issue happens with the Bar Braw Module enabled and the bar limits turned off there as well, so I beleive that the value is being reset somewhere else.

**Discord contact info**
Bolts#9418

Grygon:
Fixed in 0.9.6

Issue #252, opened 2021-07-15
[BUG] Overkill is always kh1


Grygon:
![image](https://user-images.githubusercontent.com/647846/125724047-855a2e93-c020-440e-9182-81f7b57698a3.png)

Make it kh{x} dynamically probably

Grygon:
Fixed in 0.9.6

Issue #216, opened 2021-06-06
[BUG] [0.9.3] The mech image import from Comp/Con no longer works


VexWerewolf:
**Describe the bug**
Importing from Comp/Con no longer pulls custom mech images and sets them as the mech's token image.

**To Reproduce**
1. Create a mech in COMP/CON with a custom mech image
2. Import this character via the import function

**Expected behavior**
The imported mech should have its associated image set as its token image.

**Platform (please complete the following information):**
 - OS: Windows 10
 - Foundry version 0.6.8
 - Lancer system version 0.9.3

**Discord contact info**
Vex#1075

VexWerewolf:
Pilots also no longer get custom images imported.

Grygon:
Fixed in 0.9.6

PR #258, opened 2021-07-18
handle new comp/con vault ids (also #216, #252, #256, #261)

SohumB:
#216 (mech/pilot image importing) (~~requires https://github.com/massif-press/lancer-data/pull/153~~ done!)
#252 (overkill) I _think_ is correct now — it always has a `khn` at most, and I _think_ has the explode and kh modifiers happen at the right times in all cases
#256 (can't do negative hp on the token bars) 
#261 (structure damage macro currently borked)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -139,10 +139,15 @@ let webpackConfig = shouldWatch => {
           test: /\.css$/i,
           use: ["style-loader", "css-loader"],
         },
+        {
+          test: /\.mjs$/i,
+          include: /node_modules/,
+          type: 'javascript/auto',
+        }
       ],
     },
     resolve: {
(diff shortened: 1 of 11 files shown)


SohumB:
The new comp/con vault ids embed the bucket information, which is great. But the information about the owner actually isn't persisted anywhere but in localStorage (https://github.com/massif-press/compcon/issues/1537), so we can't prepopulate from the list of users anymore, unless the user is trying to fetch a pilot that's actually in their s3 bucket.

SohumB:
Good news! the S3 bucket now contains a user's linked pilots again, so all we have to do now is actually fetch from the owner's bucket instead of the logged-in-user's. The biggest benefit there is not having to pull in the amplify graphql API.

PR #273, opened 2021-08-09
Add foundry-vtt-types as a dependency

BoltsJ:
Switches to the league types repository rather than using the very poorly aged fork of foundry-pc-types. Makes the necessary adjustments to the build configuration.

This is fairly initial support and there's likely many remaining problems, however the benefits moving forward are likely to be great.

One of the largest changes is to use declaration merging for actor and item data, rather than generics. To this end, while the various alias types haven't been outright removed yet, they now resolve simply to LancerActor and LancerItem.


src/module/actor/lancer-actor.ts
```diff
@@ -91,39 +95,67 @@ interface BoundedValue {
...
-        evasion: number;
-        edef: number;
-        save_target: number;
-        speed: number;
-        armor: number;
-        // todo - bonuses and stuff. How to allow for accuracy?
-      };
-    };
+interface DerivedProperties<T extends EntryType> {
+  // These are all derived and populated by MM
```
SohumB, line 99: Are these populated by MM specifically, or by code we use that reads from mm? If the former, is there an MM type we could use instead of listing them out?

src/module/actor/lancer-actor.ts
```diff
@@ -838,17 +857,17 @@ export class LancerActor<T extends LancerActorType> extends Actor {
...
     }
     this.data.data.derived = dr;
 
-    // Update our known values now, synchronously. 
-    dr.current_hp.value = this.data.data.current_hp
-    if(this.data.type != EntryType.PILOT) {
-      let md = this.data.data as RegEntryTypes<EntryType.MECH | EntryType.NPC | EntryType.DEPLOYABLE>;
+    // Update our known values now, synchronously.
+    dr.current_hp.value = this.data.data.current_hp;
+    if (this.data.type != EntryType.PILOT) {
```
SohumB, line 866: `if (!this.is_pilot())`, perhaps? there's a few more of these

src/module/actor/lancer-actor.ts
```diff
@@ -1119,19 +1136,22 @@ export class LancerActor<T extends LancerActorType> extends Actor {
...
+  }
+  is_mech(): this is LancerActor & { data: { type: EntryType.MECH } } {
+    return this.data.type === EntryType.MECH;
+  }
+  is_npc(): this is LancerActor & { data: { type: EntryType.NPC } } {
+    return this.data.type === EntryType.NPC;
+  }
+  is_deployable(): this is LancerActor & { data: { type: EntryType.DEPLOYABLE } } {
+    return this.data.type === EntryType.DEPLOYABLE;
+  }
```
SohumB, line 1152: I still think these type guards should state the narrower types `this is LancerActor & { data: LancerActorDataProperties<EntryType.PILOT> }` and friends.

src/module/macros.ts
```diff
@@ -1251,14 +1226,14 @@ async function rollOverchargeMacro(actor: Actor, data: LancerOverchargeMacroData
 
 export async function prepareChargeMacro(a: string) {
   // Determine which Actor to speak as
-  let mech: LancerActor<EntryType.MECH> | null = getMacroSpeaker(a);
+  let mech = getMacroSpeaker(a);
   if (!mech) return;
   const ent = mech.data.data.derived.mm;
+  // TODO: Is this only supposed to run for NPCs?
```
SohumB, line 1232: I don't believe PCs can have recharging systems, right?

SohumB (commented):
In a branch like this, it could be useful to generate a diff on the compiled output as well (unminified, ofc). I'll see if I can pull that together tomorrow.

src/module/actor/lancer-actor.ts
BoltsJ: If this is that important to you you can write the patch yourself, but as I've already explained, intersecting with that subset of `Actor.data` is no more specific than the subset I'm using here.

src/module/actor/lancer-actor.ts
BoltsJ: Looks like I missed that on. I ran a ripgrep for the pattern, but I'll get that fixed.

src/module/actor/lancer-actor.ts
BoltsJ: This was copied (including the comment) from within the previous verson of  Actor.  @whitespine would probably have a better idea.

src/module/macros.ts
BoltsJ: Is that what this is for? If so I'll just typeguard it to `this.is_npc()`

src/module/actor/lancer-actor.ts
SohumB: Gotcha. We can leave it be for now then probably.

src/module/actor/lancer-actor.ts
BoltsJ: I believe I got them all.

src/module/actor/lancer-actor.ts
SohumB: And as I've explained, what types are narrower/more specific than others is a property that is independent of what a type environment infers. If you want to make me write the patch, I will; this just seems like a weird hill to die on.

BoltsJ: I'm not continuing this conversation.

PR #283, opened 2021-08-20
Targeting automation utilities

BoltsJ:
Provide pieces needed for automating (hopefully optionally) the targeting step of attack macros.

- [x] Calculate targeted units based on a MeasuredTemplate
- [x] Add a promise the WeaponRangeTemplate workflow to allow tracking progress
- [x] Exclude centered token in Bursts
~~- [ ] Wait for targeting in macro (may not be needed)~~
~~- [ ] Allow deploying a template as part of the macro~~
~~- [ ] Remove the template macro compendium~~

Depends on #281.


BoltsJ:
Second half will go into 1.1

PR #304, opened 2021-09-02
Fix spotter

### LINKED ISSUE #299: [BUG] Spotter - Inconsistent Spotter Checkbox when moving mechs
**Describe the bug**
When operating with a Spotter-provider and Spotter-receiver, while the attack Accuracy/Difficulty box is up, if the Spotter-provider or Spotter-receiver mech either:
- Starts adjacent to the other mech and moves more than 1 space away OR
- Starts more than 1 space away and moves adjacent to the other mech

Then the visual Spotter checkbox sometimes doesn't update to reflect the change in adjacency (**_Spotter still applies to ✅ Roll, this is merely a visual bug in the Acc/Diff box_**).

**To Reproduce**
1. Create 2 Size 1 mechs: 1 with a Spotter pilot and 1 with a non-spotter pilot
2. Create a token for each of those mechs, and 1 for an NPC to target.
3. Apply Lock-On condition to the NPC target.
4. Initiate an attack against the NPC target, making sure "Consume Lock-On" is enabled
5. Move the mechs adjacent to each other, then away from each other by more than 1 space away, repeating as necessary (I tested primarily 2 and 3 spaces away).
6. Note that the Spotter checkbox sometimes doesn't update to reflect the new adjacency state (e.g. "enabled" when mechs aren't adjacent or "disabled" when they are
7. BONUS: Verify that Spotter is still applied on adjacency (or not applied when not adjacent) when ✅ Roll is clicked, despite what the Spotter checkbox indicates.

**Expected behavior**
Spotter checkbox accurately reflects adjacency after every token position update.

**Screenshots**
Using "A Mech"'s Segment Knife against "An Aegis" while adjacent to the Spotter mech (Note "Spotter" _appears_ disabled)
image

Using "A Mech"'s Segment Knife against "An Aegis" while NOT adjacent to the Spotter mech (Note "Spotter" _appears_ enabled)
image

 - OS: Windows 10
 - Foundry version 0.8.9
 - Lancer system version 1.0.0

**Additional context**
Spotter adjacency updates when other tokens move as well (so if the 2 mechs are Spotter-adjacent, but the Spotter Checkbox is empty, if I move any other token, the Spotter Checkbox will fill in to properly reflect the adjacency).  EDIT: Other "updates" (like adding/removing a condition from any token) also refresh the box.  This leads me to believe that the checkbox display may just be an async issue (one that thankfully doesn't affect the functionality).

All in all, this is a super minor bug that only really affects visual clarity of Spotter in the edge case where the Spotter-related mechs are moved while the Attack prompt is still visible.  The Spotter functionality otherwise appears to function as expected for Size 1 mechs.

**Discord contact info**
Valkyrion#6196

SohumB:
Fixes #299, and fixes spotter calcs by using Bolts' getOccupiedSpaces. Downstream of #302.

### DIFF
--- a/src/module/token.ts
+++ b/src/module/token.ts
@@ -34,36 +34,38 @@ export class LancerTokenDocument extends TokenDocument {
 export class LancerToken extends Token {
   constructor(document: LancerTokenDocument) {
     super(document);
-    this._spaces = new Set<Point>();
+    this._spaces = [];
   }
 
   /**
    * Cached occupied spaces
    */
-  protected _spaces: Set<Point>;
+  protected _spaces: Point[];
 
   /**
    * Returns a Set of Points corresponding to the grid space center points that
    * the token occupies.
    */
-  getOccupiedSpaces(): Set<Point> {
-    if (this._spaces.size === 0 && canvas?.grid?.type !== CONST.GRID_TYPES.GRIDLESS) {
+  getOccupiedSpaces(mode: "current position" | "updated position" = "current position"): Point[] {
+    let pos: { x: number, y: number } = mode == "current position" ? this.position : this.data;
+
+    if (this._spaces.length === 0 && canvas?.grid?.type !== CONST.GRID_TYPES.GRIDLESS) {
       let hitBox: PIXI.IHitArea;
       if (this.hitArea instanceof PIXI.Polygon) {
         let poly_points: number[] = [];
         for (let i = 0; i < this.hitArea.points.length - 1; i += 2) {
-          poly_points.push(this.hitArea.points[i] + this.position.x, this.hitArea.points[i + 1] + this.position.y);
+          poly_points.push(this.hitArea.points[i] + pos.x, this.hitArea.points[i + 1] + pos.y);
         }
         hitBox = new PIXI.Polygon(poly_points);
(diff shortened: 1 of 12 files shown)


src/module/token.ts
```diff
@@ -34,36 +34,38 @@ export class LancerTokenDocument extends TokenDocument {
...
-          poly_points.push(this.hitArea.points[i] + this.position.x, this.hitArea.points[i + 1] + this.position.y);
+          poly_points.push(this.hitArea.points[i] + pos.x, this.hitArea.points[i + 1] + pos.y);
         }
         hitBox = new PIXI.Polygon(poly_points);
       } else if (this.hitArea instanceof PIXI.Rectangle) {
-        hitBox = new PIXI.Rectangle(this.position.x, this.position.y, this.hitArea.width, this.hitArea.height);
+        hitBox = new PIXI.Rectangle(pos.x, pos.y, this.hitArea.width, this.hitArea.height);
       } else {
         // TODO, handle all possible hitArea shapes
         return this._spaces;
```
BoltsJ, line 64: This should probably be `return Array.from(this._spaces)` to avoid consumers accidentally mutating the cache.  Same with the final return.

SohumB: Good point!

PR #312, opened 2021-09-04
Implement RetrogradeMinis Mech Images

Grygon:
On top of #308.

If a Mech or NPC is using the default image for their current Frame/Class, turns the actor/token image into the mapped image if we know it.

TODO: Make this work on import too


Grygon:
NPC Preview
Animation

PC Preview
Animation (1)

Credits callout:
image

Grygon:
Re-drafting, tokens aren't getting their images set properly and it'll be a day or two before I can pick this back up.

src/module/actor/lancer-actor.ts
```diff
@@ -655,22 +659,33 @@ export class LancerActor extends Actor {
...
-          // TODO: Retrogrades
+          flags.top_level_data["token.img"] = new_img;
+          flags.top_level_data["token.name"] = data.callsign;
+          flags.top_level_data["token.disposition"] = this.data.token.disposition;
+          flags.top_level_data["token.actorLink"] = true;
+
+          await mech.writeback();
+
+          // If we've got a frame (which we should) check for setting Retrograde image
+          // Also check that we don't have a custom image, which would be on imgur. If so, preserve it.
```
SohumB, line 684: are all custom images on imgur? Is that a thing we should be assuming, even if it is currently true?

src/module/actor/lancer-actor.ts
```diff
@@ -30,6 +31,9 @@ import type { RegEntry, MechWeapon, NpcFeature } from "machine-mind";
 import { StabOptions1, StabOptions2 } from "../enums";
 import { fix_modify_token_attribute } from "../token";
 import type { ActionData } from "../action";
+import { frameToPath } from "./retrograde-map";
+import { string } from "io-ts";
+import { NpcClass } from 'machine-mind';
```
SohumB, line 36: You don't appear to be actually using io-ts/`string`, and the import from mm can be an `import type`.

src/module/actor/lancer-actor.ts
```diff
@@ -872,26 +891,66 @@ export class LancerActor extends Actor {
...
+                [game.system.id]: {
+                  mm_size: size,
+                }
+              },
+            });
+          }
+
+          // And copying Bolts' work, do the same thing for the token image 
+          // with Retrograde's work
+          const cached_frame = this.data.token.flags[game.system.id]?.mm_size;
```
SohumB, line 923: tsserver tells me this `cached_frame` is never used?

src/module/actor/lancer-actor.ts
```diff
@@ -1110,6 +1169,53 @@ export class LancerActor extends Actor {
...
+      robot.Flags.top_level_data["token.img"] = newFramePath;
+      changed = true;
+    }
+
+    let a = undefined;
+    let b = undefined;
+
+    if (changed) {
+      console.log(`${lp} Automatically updating image: `, newData);
+      a = await this.update(newData);
```
SohumB, line 1214: Just `await this.update(newData)` if you just want to block on it. Also lets you kill a/b.

SohumB (approved):

src/module/actor/lancer-actor.ts
BoltsJ: For comp/con imports, yes. When you upload a picture to C/C it uploads it to imgur and hotlinks it. I think the only way to bypass that would be to edit the pilot JSON directly.

src/module/actor/lancer-actor.ts
BoltsJ: Find references only shows this location for me.

BoltsJ: Also, the definitions for Token flags are at the bottom of token.ts.

src/module/actor/lancer-actor.ts
```diff
@@ -1110,6 +1169,53 @@ export class LancerActor extends Actor {
...
+   * we detect that the image isn't custom. Will check each individually
+   * @param robot     A MM Mech or NPC, passed through to avoid data overwrites 
+   * @param oldFrame  Old Frame or NPC Class
+   * @param newFrame  New Frame or NPC Class
+   * @returns         The newFrame if any updates were performed
+   */
+  async swapFrameImage(robot: Mech | Npc, oldFrame: Frame | NpcClass | null, newFrame: Frame | NpcClass): Promise<string> {
+    let oldFramePath = frameToPath[oldFrame?.Name || ""];
+    let newFramePath = frameToPath[newFrame?.Name || ""];
+    let defaultImg = robot.Type == EntryType.MECH ? "systems/lancer/assets/icons/mech.svg" : "systems/lancer/assets/icons/npc_class.svg";
```
BoltsJ, line 1185: Can use the `is_reg_mech(robot)` type guard.

src/module/actor/lancer-actor.ts
```diff
@@ -1110,6 +1169,53 @@ export class LancerActor extends Actor {
...
+    
+    if(!newFramePath) newFramePath = defaultImg;
+    let changed = false;
+    let newData: any = {}
```
BoltsJ, line 1189: This can be `Parameters<this["update"]>[0]` instead of `any` to get better type checking.

src/module/actor/lancer-actor.ts
Grygon: Correct--we'll need to patch if they change that, but couldn't think of a good catchall beyond the reverse of ! "compcon-image-assets" which would similarly need a patch if changed

src/module/actor/lancer-actor.ts
Grygon: Oops, yeah... left part of an old approach in.

src/module/actor/lancer-actor.ts
Grygon: Ooh very nice

src/module/actor/lancer-actor.ts
Grygon: And... forgot I left that in for debugging

src/module/actor/retrograde-map.ts
```diff
@@ -0,0 +1,72 @@
...
+    "BERSERKER": "systems/lancer/assets/retrograde-minis/Retrograde-Minis-Corpro-BERSERKER.png",
+    "BOMBARD": "systems/lancer/assets/retrograde-minis/Retrograde-Minis-Corpro-BOMBARD.png",
+    "BREACHER": "systems/lancer/assets/retrograde-minis/Retrograde-Minis-Corpro-BREACHER.png",
+    "CATAPHRACT": "systems/lancer/assets/retrograde-minis/Retrograde-Minis-Corpro-CATAPHRACT.png",
+    "DEMOLISHER": "systems/lancer/assets/retrograde-minis/Retrograde-Minis-Corpro-DEMOLISHER.png",
+    "ENGINEER": "systems/lancer/assets/retrograde-minis/Retrograde-Minis-Corpro-ENGINEER.png",
+    "GOLIATH": "systems/lancer/assets/retrograde-minis/Retrograde-Minis-Corpro-GOLIATH.png",
+    "HIVE": "systems/lancer/assets/retrograde-minis/Retrograde-Minis-Corpro-HIVE.png",
+    "HORNET": "systems/lancer/assets/retrograde-minis/Retrograde-Minis-Corpro-HORNET.png",
+    "HUMAN": "systems/lancer/assets/retrograde-minis/Retrograde-Minis-Misc-INFANTRY.png",
```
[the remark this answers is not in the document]
Grygon: Wasn't sure if that was the case--done!

src/module/actor/retrograde-map.ts
```diff
@@ -0,0 +1,72 @@
+export const frameToPath: Record<string, string> = {
+    "BALOR": "systems/lancer/assets/retrograde-minis/Retrograde-Minis-Horus-BALOR.png",
```
Eranziel, line 2: Noticed a few missing here that we have images for:
* Atlas
* Caliban
* White Witch
* Emperor

src/module/actor/lancer-actor.ts
Eranziel: An idea that's been kicking around in my head this morning - would it save much work if you expanded `TypeIcon()` to find the frame/class image, defaulting to the current SVG's if we don't have a Retrograde image? I think that would make it work on CC import as well.

PR #314, opened 2021-09-07
Reuse existing roll for critical hits

BoltsJ:
When creating critical hit rolls, instead of rolling an all new damage roll,
take the existing roll and smoosh it into a second new roll, then keep the
appropriate number of dice.

Apparently there's no existing issue for this one.

### DIFF
--- a/src/module/helpers/actor.ts
+++ b/src/module/helpers/actor.ts
@@ -5,6 +5,7 @@ import { simple_mm_ref } from "./refs";
 import { encodeMacroData } from "../macros";
 import type { ActionType } from "../action";
 import { LANCER } from "../config";
+import type { LancerActor } from "../actor/lancer-actor";
 // ---------------------------------------
 // Some simple stat editing thingies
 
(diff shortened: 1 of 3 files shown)


src/module/macros.ts
```diff
@@ -396,6 +395,7 @@ export async function renderMacroTemplate(actor: LancerActor | undefined, templa
     terms: [],
     result: "",
   };
+  /*
```
SohumB, line 398: I'd rather just delete the code; it'll be in git history if we need it

src/module/helpers/actor.ts
```diff
@@ -130,8 +131,8 @@ export function clicker_stat_card(
 ): string {
   let button = "";
   // Determine whether this is an unlinked token, so we can encode the correct id for the macro.
-  const r_actor = options.data.root.actor;
-  let id = r_actor.token && !r_actor.token.isLinked ? r_actor.token.id : r_actor.actor.id!;
+  const r_actor = options.data.root.actor as LancerActor;
+  let id = r_actor.token && !r_actor.token.isLinked ? r_actor.token.id : r_actor.id!;
```
SohumB, line 135: I assume this is for an unrelated bugfix?

SohumB (approved):

src/module/macros.ts
BoltsJ: Whoops, I meant to. Fixed now

src/module/helpers/actor.ts
BoltsJ: It was breaking linked sheets for me, so I had fix this just to get it tested.

PR #315, opened 2021-09-07
Inactive mechs

### LINKED ISSUE #292: [FEATURE] Add inactive mech management to pilot sheet
**Is your feature request related to a problem? Please describe.**
Currently to swap in/out your active mech, you need to be able to drag/drop from the sidebar which isn't necessarily a given. Also, it's messy.

**Describe the solution you'd like**
On import we have the necessary info to create an InactiveMechs array on the Pilot--so we should do so. Once that's populated, it would be pretty trivial to add the necessary UI around that to let players manage that on sheets.
Drop can add a mech to InactiveMechs if it isn't already present
@whitespine sound feasible for MM?

Grygon:
Accidentally built this on top of #312 

Closes #292 once complete.

Still have quite a bit of UI to mock up, but the core functionality is there


Grygon:
To do from conversation in #292: Swap our method of accessing inactive mechs by changing base pilotSheet data in getData

PR #318, opened 2021-09-08
Fix the cache invalidation strategy for getOccupiedSpaces()

BoltsJ:
Since _onUpdate seems to run before prepareData, and vision is
calculated before prepareData, sensor-sight exposed a race condition
that would cause tokens to cache their old occupied spaces when moving.
This now stores the floored coordinate and compares it the the current
coord and invalidates the cache if they differ in the getOccupiedSpaces
method directly.

### DIFF
--- a/src/module/token.ts
+++ b/src/module/token.ts
@@ -49,10 +49,9 @@ export class LancerToken extends Token {
    * Returns a Set of Points corresponding to the grid space center points that
    * the token occupies.
    */
-  getOccupiedSpaces(mode: "current position" | "updated position" = "current position"): Point[] {
-    let pos: { x: number; y: number } = mode == "current position" ? this.position : this.data;
+  getOccupiedSpaces(): Point[] {
+    let pos: { x: number; y: number } = this.data;
     // Invalidate the cache if the position is different than when it was last calculated.
-    // INFO: Multiple calls to this with different modes in the same stackframe could very rarely lead to an invalid cache.
     if (Math.floor(pos.x) !== Math.floor(this._spaces.at.x) || Math.floor(pos.y) !== Math.floor(this._spaces.at.y)) {
       this._spaces.at = { ...pos };
       this._spaces.spaces = [];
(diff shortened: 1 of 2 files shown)


src/module/token.ts
```diff
@@ -34,22 +34,30 @@ export class LancerTokenDocument extends TokenDocument {
...
 
   /**
    * Returns a Set of Points corresponding to the grid space center points that
    * the token occupies.
    */
   getOccupiedSpaces(mode: "current position" | "updated position" = "current position"): Point[] {
-    let pos: { x: number, y: number } = mode == "current position" ? this.position : this.data;
+    let pos: { x: number; y: number } = mode == "current position" ? this.position : this.data;
+    // Invalidate the cache if the position is different than when it was last calculated.
+    if (Math.floor(pos.x) !== Math.floor(this._spaces.at.x) || Math.floor(pos.y) !== Math.floor(this._spaces.at.y)) {
```
SohumB, line 55: It concerns me that we change what we're looking at to invalidate the cache depending on what the `mode` sent in is. It's probably fine, overall — in any case where something requests the updated position, that position probably _is_ the current position on a later call, but it makes me uncomfortable.

I'm not requesting any code changes here, but can we add like a rust-style "this unsafe block is safe, we think, because" comment?

SohumB (approved):

src/module/token.ts
BoltsJ: Even if a later call requests the current position, the result will be the same if the coordinate for pos is the same. I'm not sure what the unsafe condition is here.

SohumB:
It's that "if", right — in theory we could flip a bunch of cache
invalidations if a bunch of requests came in in one synchronous stackframe
for updated/current/updated/current. I didn't mean to suggest anything this
code is doing is *unsafe* (ultimately the worst that can happen is we do
more computation than necessary), I was just making an analogy to the kinds
of doc-comments you see sometimes in Rust.

BoltsJ:
Ah, that makes sense. Although, I wonder if we even need the mode switch anymore. `this.position` and the coords in data will be the same for a stationary token.

PR #323, opened 2021-09-11
Create an object setting to hold automation config

BoltsJ:
Moves the various automation settings to a single config object and adds
a configuration form for the settings options. Adjusts the existing
consumers of the old settings to check the new object. 

 - [x] Create the new config
 - [x] Create a config form
 - [x] Migrate existing configs
 - [x] Adjust consumers to use the new setting
 - [x] Adjust automation on some macros to reflect updates in descriptions
 - [x] BONUS: Svelte dialog for structure/stress prompts
   - [x] Add dialog to sliding HUD zone
   - [x] Make it look pretty
   - [x] Reconfigure structure/stress macro


BoltsJ:
image

BoltsJ:
image
Preview of condition reference on chat message. The name of the status is now a link to the compendium status item.

BoltsJ:
image
Preview of confirmation dialogs.

src/module/actor/lancer-actor.ts
```diff
@@ -243,17 +252,27 @@ export class LancerActor extends Actor {
...
-      this.structureMech();
-    } else {
+    if (!this.is_mech() && !this.is_npc()) {
       ui.notifications!.warn("Can only structure NPCs and Mechs");
       return;
     }
+    const ent = await this.data.data.derived.mm_promise;
+    if (getAutomationOptions().structure && !reroll_data) {
+      if (ent.CurrentHP < 1 && ent.CurrentStructure >= 0) {
+        ent.CurrentHP += ent.MaxHP;
```
Eranziel, line 264: A slight bug here shown by the macro firing automatically - if `ent.CurrentStructure == 1` then we shouldn't add max HP, because the guy's dead. Similar for overheat, I think we shouldn't subtract heat if `ent.CurrentStress == 1`.

src/module/helpers/struct_stress/Form.svelte
```diff
@@ -0,0 +1,79 @@
...
+
+  function getIcon(kind: "structure" | "stress") {
+    if (kind === "stress") return "reactor";
+    return kind;
+  }
+
+  function getDamage(a: LancerActor) {
+    if (!a.is_mech() && !a.is_npc()) return 0;
+    return a.data.data.derived[stat].max - a.data.data.derived[stat].value + 1;
+  }
```
SohumB, line 25: This is fine, but the sveltier way to do it, given that these functions are only ever called with the current kind/actor, would be to use reactive variables.

```typescript
$: icon = stat === "stress" ? "reactor" : stat;
$: damage = getDamage(lancerActor);
```

```html
...
<i class="cci cci-{icon} i--m i--light" />
...
Roll {damage}d6 to determine what happens.
```

PR #325, opened 2021-09-12
Add isAttack flag to MeasuredTemplates created by the accdiff dialog

BoltsJ:
Allows external integrations to filter what the source of a
MeasuredTemplate is in order to, for example, remove attack templates at
the end of a turn.


BoltsJ:
Rolled into #323

PR #330, opened 2021-09-15
feat(macro):streamline struct & stress

msprijatelj:
# Description

This is a minor PR with the goal of streamlining the code behind the Structure and Stress macros.

- [x] Current vs Max Struct/Stress check extends to Struct/Stress values over the maximum (since it's possible to set Struct/Stress higher than the maximum)
- [x] Moves the `ent.writeback()` for Struct/Stress prior to the `>= MaxStress` check to prevent a race condition that trips up CUB Trigger automation
- [x] Passes the current actor to `renderMacroTemplate` to bypass the need to get the current speaker (possibly improving performance? let me know if I misunderstood the previous code, thanks)

Feedback is welcome.  This is my first PR on the repository, and Javascript isn't my first language, so **_please_** let me know if I missed something.  I want to learn and improve so I can make meaningful contributions to the project!


Grygon (approved):

Eranziel (approved):

BoltsJ (approved):
LGTM

PR #355, opened 2021-10-03
Integrate lancer-initiative into the system

### LINKED ISSUE #5: [FEATURE] Initiative System
Implement Lancer-specific combat initiative logic.

BoltsJ:
Adds lancer-initiative from npm and uses LancerCombat, LancerCombatant
and LancerCombatTracker to provide the same functionality as
lancer-initiative. Provides a customized CombatTrackerConfig that adds
lancer-initiative settings to that form. The same appearance config form
from lancer-initiative is included and can be accessed from the
CombatTrackerConfig.

As part of this, the same Hooks provided by lancer-initiative are
provided, which allows any extension (li-player-request-extension for
example) for lancer-intiative to work with the system as well. Those
hooks are "LancerCombatRequestActivate" which fires when a non-GM clicks
the activation button, and "LancerInitiativeInit" which fires after the
settings and CONFIG object have been set up.

Fixes #5


BoltsJ:
Also, the current version of lancer-intitiative will simply disable itself and display a warning if it detects it's running on a system that already provides it.

PR #370, opened 2022-01-09
Rearranging `macros.ts` into individual files.

mason-bially:
This is a functionality preserving rearrangement of the macro code. No code was changed yet, just moved into 21 individual files. This makes it substantially easier to see what macro code interacts with which other pieces of macro code. Though `macros/attack.ts` is still a very large group of code (at almost 600 lines long) that could likely use some more rearranging.

I did this with individual commits to make it easier to track and verify, though that last one is a doozy.

A number of thoughts on further organizational work:
- Two files are prefixed with an underscore, `_hotbar.ts` and `_template.ts` both seemed out of place here.
- The `util.ts` and `render.ts` are used quite extensively and are not directly game concepts, and might also be in need of an underscore to separate them from the otherwise game-conceptual names in the folder.
- it might be worth considering renaming the folder to `macro`.
- Significant parts of `interface.ts` seem to be used for this folder.
- Some macro names are poorly chosen:
    - `prepareOverheatMacro` should probably be `prepareStressMacro` to follow with `prepareStructureMacro` (or `prepareStructureMacro` should be renamed `prepareStructureDamageMacro`).
    - `prepareChargeMacro` should probably be `prepareNpcRechargeMacro` or similar to make it less ambiguous.
    - `stabilizeMacro` and `fullRepairMacro` seem strangely named, though I note they are also ones directly exposed to the user. Perhaps `userStabilizeMacro` and `userFullRepairMacro`?


src/module/macros/_hotbar.ts
```diff
@@ -0,0 +1,147 @@
...
+        }
+        break;
+    }
+
+    // TODO: Figure out if I am really going down this route and, if so, switch to a switch
+    if (data.type === "actor") {
+      title = data.title;
+    } else if (data.type === "pilot_weapon") {
+      // Talent are the only ones (I think??) that we need to name specially
+      if (data.type === EntryType.TALENT) {
```
Eranziel, line 118: Note for future - this is weird. Why are talents inside pilot weapons? Seems unreachable.

src/module/macros/activation.ts
```diff
@@ -0,0 +1,168 @@
...
+}
+
+async function _prepareTextActionMacro(
+  actorEnt: Mech | Pilot | Npc,
+  itemEnt: Talent | MechSystem | NpcFeature,
+  index: number
+) {
+  // Support this later...
+  // TODO: pilot gear and NPC features
+  if (itemEnt.Type !== EntryType.MECH_SYSTEM && itemEnt.Type !== EntryType.TALENT) return;
```
Eranziel, line 108: Note for future - I think we have some `is_reg_x()` helper functions for uses like this. Unless we only have those for the actor types.

SohumB:
Looks good to me. I went digging for any edge cases in the compiled output, but it all looks just like code movement, as you'd expect. 

I'd support all your macro renamings — but for public macros, the old names would need to be kept around but log a deprecation message. If you feel like doing that as well, great, otherwise I'm happy to merge this.

src/module/macros/overcharge.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  // And here too... we should probably revisit our type definitions...
+  let rollText = actor.getOverchargeRoll();
+  if (!rollText) {
+    ui.notifications!.warn(`Error in getting overcharge roll...`);
+    return;
+  }
+
+  // Prep data
+  let roll = await new Roll(rollText).evaluate({ async: true });
```
Eranziel, line 30: My gut questions why we need to `evaluate` here.

mason-bially:
> I'd support all your macro renamings — but for public macros, the old names would need to be kept around but log a deprecation message. If you feel like doing that as well, great, otherwise I'm happy to merge this.

I more put down the thoughts for posterity's sake. Might be worth having a meta issue for "renames we would like to do" to keep track of for pending breaking changes and manage all the depreciation at once so we keep them on track. Either way if I was to do the renames I would want them separate from this PR anyway.

src/module/macros/overcharge.ts
```diff
@@ -0,0 +1,66 @@
...
+  if (getAutomationOptions().overcharge_heat) {
+    mech.CurrentHeat = mech.CurrentHeat + roll.total!;
+  }
+
+  await mech.writeback();
+
+  return rollOverchargeMacro(actor, mData);
+}
+
+async function rollOverchargeMacro(actor: LancerActor, data: LancerOverchargeMacroData) {
```
Eranziel, line 52: Naming and functionality is counterintuitive here. `prepareOverchargeMacro` does the roll, `rollOverchargeMacro` renders it.

src/module/macros/full-repair.ts
```diff
@@ -0,0 +1,42 @@
+// Import TypeScript modules
+import { LANCER } from "../config";
+import { getMacroSpeaker } from "./util"
+import { prepareTextMacro } from "./text"
+
+const lp = LANCER.log_prefix;
+
+export function fullRepairMacro(a: string) {
```
Eranziel, line 8: Maybe full repair, short rest, and stabilize  belong in the same file?

src/module/macros/attack.ts
```diff
@@ -0,0 +1,593 @@
...
+        usedLockOn: tad.usingLockOn,
+      };
+    }),
+  };
+}
+
+
+type AttackMacroOptions = {
+  accBonus: number;
+  damBonus: { type: DamageType; val: number };
```
Eranziel, line 81: Note to self - where do we use this? Shouldn't `damBonus.val` be a string, since it's usually "1d6"?

src/module/macros/attack.ts
```diff
@@ -0,0 +1,593 @@
...
+  let rerollMacro = {
+    title: "Reroll attack",
+    fn: "prepareEncodedAttackMacro",
+    args: [actor.data.data.derived.mm!.as_ref(), item.id, options, promptedData.toObject()],
+  };
+
+  await rollAttackMacro(actor, atkRolls, mData, rerollMacro);
+}
+
+export async function openBasicAttack(rerollData?: AccDiffData) {
```
Eranziel, line 299: A possibility for downsizing this file further would be separating the handlers for the attack HUD. @SohumB, thoughts?

At a glance, that would include `checkTargets`, `openBasicAttack`, `applyPluginsToRoll`, and maybe `attackRolls`.

Eranziel (commented):
Writing down some thoughts on the state of things from mobile.

Eranziel:
I'm on board with your proposed renames as well, and I agree we should soft-deprecate the old names in the next release to hard-deprecate them later. I'll open an issue for tracking those.

src/module/macros/attack.ts
SohumB: I'm pretty sure this is for flat damage bonuses, not the actual damage roll. That said, I had a quick look and it seems we only have code for consuming this option, and nothing that generates it.

src/module/macros/attack.ts
SohumB: Fine by me, though it is difficult to figure out a clean separation here between "attack-related things" and "things the attack HUD uses". I'd suggest getting it down to a 600-line file is a great first step, and we can reorg that file the next time someone needs to actually touch it.

src/module/macros/_hotbar.ts
mason-bially: I removed this entire branch system in #375

Issue #374, opened 2022-01-09
[REFACTOR] Macro Cleanup and De-duplication


mason-bially:
This is a tracking issue for macro cleanup and de-duplication that can be done to improve macro code readability and understand-ability.

As an example of what this is aimed to fix:
- `prepareActivationMacro` dispatches from chips
  - for systems, is partially responsible for handling charges 
  - for most things this is just displaying text
- `prepareItemMacro` dispatches from the item sheets
  - does not handle charges at all 
  - for systems invokes `rollSystemMacro`
    - does not handle charges
    - dispatches into displaying text
  - failure to dispatch goes to displaying text

Related issues:
- #371 An issue for tracking macro renames and their depreciation.
- #370 what originally kicked this off, rearranging the macros into files for improved readability.

mason-bially:
I think most of the `*MacroData` interfaces could be moved into their relevant macro files. Most of them are never invoked outside of their `prepare*Macro` / `roll*Macro` pairs, and when they *are* those pairs are being imported anyway.

mason-bially:
Okay so NPCs don't *have* activation chips. They have tags that say "quick action" (and tags can be dragged, though we should probably reject putting them on the hot bar).

mason-bially:
So currently the `prepareItemMacro` leads to some weird behavior on the hot bar. Consider this scenario:

- I drag an item T from actor A's sheet to the hot bar.
- Scenario: I click the hotbar with actor A selected
    - Outcome: Actor A performs the action.
- Scenario: I click the hotbar with actor B selected, who also has the item.
    - Outcome: Actor B performs the action.
- Scenario: I click the hotbar with actor C selected, who does not have the item.
    - Outcome: Actor A performs the action.
- Scenario: I click the hotbar with no actor selected.
    - Outcome: Actor A performs the action.

I think the last two cases are in need of massaging.

mason-bially:
So where I am getting to is limiting the space of macros to this:
- `prepareActivationMacro` for activation chips and calling specific actions on things
- `prepareItemMacro` for item references directly
    - though I may want to interdict this / rename the macro for things like traits and talents which have to index into an "item" to display properly (which, btw, we need a cleaner interface for). I'm thinking `prepareDisplayMacro`
        - this is also what dragging the little "message" button will make as opposed to the "item box" (which is also used for building sheet content)
    - the reason for separating this is that "display" is a fixed kind of macro where as "item" is a more general *contextual* macro.
[...]

PR #375, opened 2022-01-09
Cleaning up dragging and dropping macros to use a consistent mechanism.

mason-bially:
As part of #374 this refactors all of the drag and drop macros with custom handler to use a single consistent mechanism.

Below is a prototypical example:
```ts
        let macroData: LancerMacroData = {
          iconPath: `systems/${game.system.id}/assets/icons/macro-icons/xyz.svg`,
          title: xyz.Name,
          fn: "prepareXyzMacro",
          args: [actor_id, xyz.id],
        };

        macro_button = `<a class="cci cci-xyz i--l i--dark xyz-macro lancer-macro" data-macro="${encodeMacroData(macroData)}"></a>`
```

This uses the `lancer-macro` hook to encode the roll when dragged. This seems the most preferable way to implement buttons that can *always* be turned into hotbar macros.

These macros are admittedly "bound" to the sheet of the actor they were dragged off of. While on the other hand dragging the "item" off the sheet continues to create "item macros". Solving the issue of how to turn "item macros" into "bound macros" or vice-versa; or ideally in time, both, to context specific macros is a later issue. For now I am simply trying to reduce the volume of code and mechanisms in play, and removing special cases where I can.

Additionally this removes a number of vestigial code paths, and fixes a number of minor consistency issues in the use and dragging of macros.


mason-bially:
(also based off of #373, since this continues to improve on that)

PR #395, opened 2022-01-27
Stabilize, Full Repair, StructStress Macro QoL changes

msprijatelj:
# Description
This change adds automated Burn clear to the Stabilize macro.  It also automatically clears the Exposed ActiveEffect, if present.

Biggest challenge I’m facing right now is figuring out how to get these automatic stat updates to play nice with CUB triggers and automated TokenMagicFX macros, though that is probably a personal mods+macros problem instead of a Lancer problem.

UPDATE: Also added Overshield + Burn clear to Full Repair macro.
UPDATE 2: Also added `remove_all_active_effects` and `remove_nontier_active_effects` methods to LancerActor (drawing from Foundry Community Macros for help); the former gets called by Full Repair (subject to change).
UPDATE 3: Limited the number of dice requested by the Struct/Stress Svelte popup to be the maximum Struct/Stress of the triggering actor.  Fixes the Svelte lockup described in #392.

## Issues Closed
Closes #391 
Closes #392

Issue #189, opened 2021-05-08
[FEATURE] Damage Application


VexxMyst:
**Is your feature request related to a problem? Please describe.**
the Mess/Mars module has an feature where you can apply damage from a roll directly to a targeted character.

**Describe the solution you'd like**
If a roll is made against a targeted character, the GM can right-click on the character section and select 1/2x, 1x, or 2x damage. If full automation is desired, the GM could instead have a button next to the hit/miss marker that applies the damage, taking resistances and Exposed into account automatically.

**Describe alternatives you've considered**
See if Moerill would be interested in supporting Lancer with Mars. As the system already has hit detection and acc/diff integrated, the module's main functions seem redundant.

**Additional context**
A mockup showing both suggestions:
![lancerhitmock](https://user-images.githubusercontent.com/74265604/117553764-62077f80-b021-11eb-9f5b-989595280c33.PNG)

**Discord contact info**
VexxMyst#6669

ScyllPoesis:
Manual damage application is in the cards, yes yes. We've discussed it internally and I have some basic designs for it, but still need to shop it out. Currently my idea is similar to MidiQOL's damage cards, which on hits will calculate what it thinks the damage should be for each target, listing out things like resistances and such, and allowing the GM to apply the suggested damage or half/double it if there are other factors it missed or aren't supported.

msprijatelj:
To chime in with my 2 cents: I'd love it if automatic damage reduction/amplification could be applied for Resistance, Exposed, Armor, Miss-for-half-damage, etc, **_without_** the need for a GM to click to verify damage.  I'm just imagining scenarios where a Monarch or a Barbarossa hits a ton of enemies at once, and the number of clicks needed to verify the damage if I need to apply damage to each target on a case-by-case basis.  Ideally, I could roll damage once and have the system automatically take care of the rest.

Eranziel:
While I agree that fully automatic would be great, the fact at the end of the day is that our data model can't account for all of Lancer's edge cases and special cases. *Especially* in cases like hitting many characters in one attack it is very likely that some of the damage would be applied incorrectly, making more work for the GM instead of less.

msprijatelj:
I am sure that full Lancer damage automation, with all edge cases covered, would require a significant amount of developer effort.  However, I am not pushing for full damage automation, here.  I would be _beyond_ happy to see the automation of the 4-step Calculating Damage algorithm from pg. 67 of the Lancer Core Book.  

[...]

PR #400, opened 2022-02-19
[FEATURE] Damage calculation for Actors

msprijatelj:
# Description
This PR adds a damage-application function to Actors for later implementation.  It assumes that incoming damage is arranged in a `Damage[]` object. From there, it condenses all damage into a single `AppliedDamage` object, which simply tracks the total incoming damage for each damage type. 

It also assumes that "Heavy Gunner"-like features occur before the function is called.  The most complicated part was getting the "armor-application" logic correct when dealing with mixed damage types (i.e. to which damage pools is armor applied *first*).  

Notable features:

- Convert Heat to Energy if the actor lacks `CurrentHeat`.
- Doubles all non-Burn, non-Heat damage if the target is exposed ("Variable" is treated as Kinetic, Energy, or Explosive-typed since it usually is not possible to make it Burn or Heat).
- Reduce damage by Armor, with carry-over if there is leftover damage.
- Armor reduction applied in order depending to favoring attackers or defenders (defenders by default, automation option to come).
- Applying resistance for all "Resistances" possessed by the actor.
- Bypassing Armor and/or Resistance with `ap` and `paracausal` parameters.
- Applying Heat if the target if they have `CurrentHeat`.
- Applying damage to Overshield first, then to HP.

### DIFF
--- a/src/module/actor/damage-calc.ts
+++ b/src/module/actor/damage-calc.ts
@@ -1,0 +1,29 @@
+import {
+  Damage,
+  DamageType,
+} from "machine-mind";
+
+export class AppliedDamage {
+  public Kinetic: number
+  public Energy: number
+  public Explosive: number
+  public Burn: number
+  public Heat: number
+  public Variable: number
+
+  public constructor(damageData: Damage[]) {
+    this.Kinetic = this.sum_damage(damageData, DamageType.Kinetic)
+    this.Energy = this.sum_damage(damageData, DamageType.Energy)
+    this.Explosive = this.sum_damage(damageData, DamageType.Explosive)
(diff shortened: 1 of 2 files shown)


msprijatelj:
This PR currently builds off of my QoL changes from #395; I could probably divorce them from this PR if needed.

mason-bially:
For posterity this is ticket #189

msprijatelj:
Pushed a small change that:
1. Uses the correct logic for `ap=true` (i.e. if `ap=true` set `armor=0`
2. Initializes `leftover_armor` as a number instead of the value of `armor` for clarity of what the variable is supposed to do (plus a relevant comment)

PR #413, opened 2022-04-16
Implement new sharecode api endpoint for pilots, fix cache logic

Kuenaimaku:
This implements the new api endpoint for getting pilot data from a share code, as the vault codes have been deprecated.

additionally, this changes how the pilot data is gathered when a pilot is selected from the sync dropdown input after having authenticated - instead of relying on the cloudOwnerID and cloudID, it recreates the key to access aws-storage based on the pilot data that's returned within pilotCache.

note that the api key is currently hard-coded, depending on the results of testing how frequently these endpoints are hit, we may have to create an api key system setting in the future.

### DIFF
--- a/src/module/compcon.ts
+++ b/src/module/compcon.ts
@@ -1,4 +1,3 @@
-import { AUDIO_FILE_EXTENSIONS } from "@league-of-foundry-developers/foundry-vtt-types/src/foundry/common/constants.mjs";
 import type { PackedPilotData } from "machine-mind";
 import type { CachedCloudPilot } from "./interfaces";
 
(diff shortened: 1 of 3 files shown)


src/module/compcon.ts
```diff
@@ -1,3 +1,4 @@
+import { AUDIO_FILE_EXTENSIONS } from "@league-of-foundry-developers/foundry-vtt-types/src/foundry/common/constants.mjs";
```
mason-bially, line 1: What's going on here?

mason-bially:
We should probably make an issue for tracking the removal of the depreciated code, there is a lot of stuff going on in there that will no longer need to in the near future.

Besides the strange import I commented on above this looks good.

src/module/compcon.ts
Kuenaimaku: That's an interesting auto-include from vscode - i'll remove it in a commit now.

Kuenaimaku:
Agreed, removing that depreciated code and simplifying the sync ui screen in the future might be for the best. Now that we don't have to keep cloudOwnerID and cloudID around, we could probably simplify it using a single string property and attempt to go to the two methods of getting a pilot from comp-con based on regex, instead of the more complex string interpolation.

Eranziel (approved):

PR #419, opened 2022-04-30
dev pulldown

ScyllPoesis:
Pulldown new version info for consistency when reviewing versioning in dev.


ScyllPoesis:
Don't want to clobber template workflows, will update version and ride along a PR.

PR #455, opened 2022-08-24
Fix roll aggregation for attack rolls

BoltsJ:
The length of critical damage roll array was not properly checked so
the aggregation code was treating every roll as a critical, even if no
critical damage rolls were included.

I'm also going to see if I can make the attack and damage rolls show
separately without needing to split the cards.

### DIFF
--- a/src/module/macros/attack.ts
+++ b/src/module/macros/attack.ts
@@ -384,6 +384,8 @@ export async function checkTargets(
         let target = targetingData.target;
         let actor = target.actor as LancerActor;
         let attack_roll = await new Roll(targetingData.roll).evaluate({ async: true });
+        // @ts-expect-error DSN options aren't typed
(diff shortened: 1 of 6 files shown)


src/module/macros/attack.ts
```diff
@@ -384,6 +384,8 @@ export async function checkTargets(
         let target = targetingData.target;
         let actor = target.actor as LancerActor;
         let attack_roll = await new Roll(targetingData.roll).evaluate({ async: true });
+        // @ts-expect-error DSN options aren't typed
+        attack_roll.dice.forEach(d => (d.options.rollOrder = 1));
```
Eranziel, line 388: Does this need a condition to check whether DSN is enabled, or does it just add an unused `rollOrder` field in that case?

BoltsJ: It's unused if dsn isn't enabled

PR #456, opened 2022-08-26
Add a setting for square grid measurement modes.

BoltsJ:
Since the rules on square grids in core are all but non-existent, it's up
to the GMs to arbitrate how diagonals work on them. This has options for
the four methods I've seen in various games.

* 1-1-1: This is the foundry default way. Diagonal movement costs 1
* 1-2-1: Diagonal movement alternates cost between 1 and 2
* 2-2-2: Manhattan distance. Diagonal movement costs 2
* Euclidian: Rounds the euclidian distance to the nearest grid space

### DIFF
--- a/src/module/settings.ts
+++ b/src/module/settings.ts
@@ -79,6 +79,20 @@ export const registerSettings = function () {
     default: false,
   });
 
+  game.settings.register(game.system.id, LANCER.setting_square_grid_diagonals, {
+    name: "lancer.squaregriddiagonals.name",
+    hint: "lancer.squaregriddiagonals.hint",
+    scope: "world",
+    config: true,
+    type: String,
+    choices: {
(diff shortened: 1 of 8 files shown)


src/module/settings.ts
```diff
@@ -79,6 +79,20 @@ export const registerSettings = function () {
...
+    config: true,
+    type: String,
+    choices: {
+      "111": "lancer.squaregriddiagonals.111",
+      "121": "lancer.squaregriddiagonals.121",
+      "222": "lancer.squaregriddiagonals.222",
+      "euc": "lancer.squaregriddiagonals.euc",
+    },
+  });
+
```
dodgepong, line 95: Does this need a `default`? I think v10 requires it but I may be misreading how that works.

BoltsJ: Yes, but I could have sworn I had it in there.

PR #459, opened 2022-09-01
[WIP] Add Support for Foundry v10

dodgepong:
Main changes so far:
* Add a few missing default settings (lancer initiative, installed LCPs)
* Fix an issue when referring to folders
* Convert usages of `(actor|item).data.data` to `(actor|item).data.system`
* Convert usages of `(actor|item).data.*` to `(actor|item).*`
* Update manifest file to new v10 format (new compatibility object, authors object, and using `id` instead of `name`, compendium pack `type` instead of `entity`)

Note that, since we're still on v8 types, I've added a pile of `@ts-expect-error` to all the document changes which should hopefully get caught when we convert to v10 types.


dodgepong:
Known issue: dragging items onto sheets current does not work.

jaha7419:
Hey, not super well versed with github. Is there a way to use this fork temporarily until the main Lancer package is updated? Thanks in advance for your help.

Eranziel:
Going to go ahead and merge this into dev, we can keep working on v10 there. Thanks for the contributions, everyone!

PR #464, opened 2022-09-04
fix(item): add delete button to range editors

msprijatelj:
This does for Range editors what #454 did for damage editors, adding a "remove" option for unwanted ranged options.


msprijatelj:
Screenshots:

NPC Feature before deletion:
image

NPC Feature after deletion:
image

Mech Weapon before deletion:
image

Mech Weapon after deletion:
image

whitespine (dismissed):

PR #470, opened 2022-09-07
Add lint-staged to automatically format staged files using prettier

BoltsJ:
Additionally, adds a prettierignore file to ignore src/aws-exports.ts
since that file is auto-generated.


Eranziel:
It's a beautiful thing when stuff just works. Thanks!

PR #478, opened 2022-09-28
Add @LancerID text enricher

BoltsJ:
Allows linking items and actors by lid allowing world portable links in
journals.

It first searches the world collections, then the packs if nothing is found.
When searching the packs, it uses the index to avoid loading the document
unnecessarily.

This could be useful for people creating adventures as it allows journal links
to be created that aren't dependent on world IDs and could be used to list npc
features in a way that allows GMs to quickly recreate npcs without having to
include paid lcp data.

Example usage: `@LancerID[mf_standard_pattern_i_everest]`, `@LancerID[mw_shotgun]{Boomstick}`


Eranziel (approved):
Looks good in testing.

PR #479, opened 2022-09-30
feat(macros): scan to journal macro added

msprijatelj:
This PR adds an alternative "Scan (Journal)" Macro that exports a Scan entry to the Journal. Usable by GMs and players with Journal Creation permissions. Thanks to Vixiea#1976 for performing most of the legwork to export Scans to the journal!


msprijatelj:
Ran into an issue with Foundry treating Folder Creation permissions as different than Journal Creation permissions, and that there doesn't seem to be a way to grant Folder Creation permissions to players! So that's fun. Fixed the macro to provide an informative warning when the scan target folder needs created by a GM.

Eranziel (approved):

PR #528, opened 2023-02-23
Update compcon-login-form.ts

tachester:
I submitted a bug mentioning the lowercase email issue; here is my simple proposed solution. Essentially, YOU make the email lowercase just before sending the form data through the network; that way the users can type in whatever casing they want and it won't be an issue.

I assume it was compcon that was forcing all emails to be lowercase, this'll make your app a little more case friendly.

(Side note: I was trying to troubleshoot this issue with my GM who is running foundry; took us forever to figure out it was a casing issue. There were no obvious feedback messages that let us know casing was important. Either that or we overlooked the heck out of it.)

### DIFF
--- a/src/module/helpers/compcon-login-form.ts
+++ b/src/module/helpers/compcon-login-form.ts
@@ -23,7 +23,9 @@ export default class CompconLoginForm extends FormApplication {
   async _updateObject(_event: any, formData: any) {
     try {
       const { Auth } = await import("@aws-amplify/auth");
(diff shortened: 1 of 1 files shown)


tachester:
Based on your response to the submitted bug,  I've changed the pull request to simply attempt to login twice; once with case sensitivity and one without.

Hopefully, I didn't do it in a way that goes against any of your team's conventions.

Eranziel (approved):

PR #531, opened 2023-03-06
feat(macros): add new Ruin Tokens macro that searches for files

msprijatelj:
# Description
This PR adds an updated version of the Ruin Tokens macro called "Ruin Tokens (File Search)". It improves upon the existing macro (~~which has been left intact, just in case of legacy users~~ which is being replaced by this PR) by searching for image files in a provided path within the FoundryVTT/Data directory. Default top-level directory is `lancer-wrecks`, with subdirectories corresponding to sizes supported by the system (`S-HALF`, `S1`, `S2`, `S3`, and `S4` by default). The macro will retrieve all filenames within a subdirectory depending on the target's size and pick one at random. If it fails to find the directory or files, it will throw a warning and not update the token image.

In addition, the macro will automatically set a "Wrecked" token's HP = Size * 10, Heat = 0, and Burn = 0.

Credit to @C-S-McFarland for providing the baseline macro code.


martinAbsolute:
Since the module already includes basic retrograde minis, what about adding the wrecked mech pngs as defaults? Would result in a much better out-of-the-box experience.

msprijatelj:
It would depend on permission from Retrograde minis, plus the point of having the `lancer-wrecks` folder external to the Lancer System itself is so it doesn't get overwritten on an update. I'd be willing to update this macro in the future provided the default Wrecks get added, but I think that's out of scope for this PR.

Eranziel:
@msprijatelj do you see any value in keeping the old Ruin macro, or should we just replace it with this one? It wouldn't be a breaking change for anyone since the old one needed to be imported and customized before it was usable anyone.

msprijatelj:
@Eranziel I don’t see a whole lot of value to keeping the old one, besides for people who already set up their wrecks and would need to redo them in the new style (and somehow need to import the old wreck macro too, like an accidental deletion). I can remove the old version next time I’m at my computer, unless you want to do that tweak yourself.

msprijatelj:
Completely replaced the old Ruin Tokens macro with this new File Search version.

msprijatelj:
@Eranziel Added a Biological tag bypass for Bio NPCs (fun fact: this already worked on Pilots, but only because they don't actually have an assigned size, so it throws a different warning lol). Please merge this before I make it run DOOM xD.

Eranziel:
Aw, but running Doom in Foundry would be cool. 😢

Eranziel (approved):

Issue #529, opened 2023-02-26
[BUG] Mech tokens reset to default image after compcon import/sync


Ward-Ceyssens:
**Describe the bug**
When syncing from compcon, the prototype token images for mechs are reset to the default despite having custom images. 

This does not happen to deployables or pilots.

**To Reproduce**
1. Have a mech with a custom token image
2. sync your pilot from compcon
3. check the prototype token on the mech

**Expected behavior**
mechs should keep customized token images when syncing

**Platform (please complete the following information):**
 - OS: Windows
 - Browser : Foundry client
 - Foundry version 10.291
 - Lancer system version 1.5.1

**Additional context**
I was bored so i did some manual bugfixing: 
it seems to originate from [swapFrameImage/line 1474](https://github.com/Eranziel/foundryvtt-lancer/blob/master/src/module/actor/lancer-actor.ts#L1474) being truthy because in [sync_mech/line 915](https://github.com/Eranziel/foundryvtt-lancer/blob/master/src/module/actor/lancer-actor.ts#L915) oldFrame gets passed as null and this.token seems to also be null. Passing mech.Loadout.Frame instead seems to fix it, but the real problem is probably this.token being null rather than oldFrame. hope that saves you some time searching.

PR #540, opened 2023-03-31
fixed issue #529 Mech tokens resetting to default image on import

Ward-Ceyssens:
I opened #529 a bit ago but recently had some time on my hands and figured I'd try fixing it myself

just changed token.img to prototypeToken.texture.src, which is the v10 path to the token image.


Eranziel (approved):
Thanks for the fix! Just in time for a little patch. ;)

PR #562, opened 2023-07-08
Changed bonds to store their system data as strings rather than indexes

whitespine:


src/module/models/shared.ts
```diff
@@ -306,16 +306,16 @@ export class SyncUUIDRefField extends fields.StringField {
...
   }
 
   /** @override */
   initialize(value: string, model: any) {
     // Expand to a somewhat reasonable range. `prepareData` functions should handle the rest
     return {
-      min: 0,
-      max: 0,
+      min: this.options?.min ?? 0,
+      max: this.options.max ?? 0,
```
Eranziel, line 318: Missing `?` after options

src/module/models/shared.ts
```diff
@@ -306,16 +306,16 @@ export class SyncUUIDRefField extends fields.StringField {
...
 // This is 10% so we can show them with bars, and 90% because usually the max is computed and we don't want to confuse anyone
 export class FakeBoundedNumberField extends fields.NumberField {
-  constructor(a: any = undefined) {
-    super(a);
+  constructor(options: object = {}) {
+    super(options);
   }
 
   /** @override */
   initialize(value: string, model: any) {
```
Eranziel, line 314: Just noticed - wouldn't `value` be a `number`?

Eranziel (approved):
Alright, I see what you're doing. Thanks for the cleanup. :+1:

Noticed a couple minor things while reading through the changes, see below.

src/module/models/shared.ts
whitespine: Yeah probably - these aren't really type checked anyways though, since the `fields` api is entirely novel to v10+ and my attempts to shim it just didn't really end up working. I'll do a lil fix though. Definitely just copy pasted something.

PR #563, opened 2023-07-09
2.0 tag fixups

whitespine:
Fixed tags not being addable now thata they aren't an item.
- Tags now addable via a + on tag card on an item
- This will make a blank tag, which can be edited / removed via the right click context menu
- Also streamlined the compact tag helper while i was at it.

### DIFF
--- a/src/module/helpers/commons.ts
+++ b/src/module/helpers/commons.ts
@@ -389,6 +389,23 @@ export function extendHelper(
   };
 }
 
+/**
+ * Use this when invoking a helper from outside a helper.
+ * A shitty hack that will break if handlebars partials are invoked
+ * @argument fake_data Will be used as the "data" for the hash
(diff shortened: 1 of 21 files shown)


src/module/helpers/commons.ts
```diff
@@ -389,6 +389,23 @@ export function extendHelper(
...
 }
 
+/**
+ * Use this when invoking a helper from outside a helper.
+ * A shitty hack that will break if handlebars partials are invoked
+ * @argument fake_data Will be used as the "data" for the hash
+ */
+export function spoofHelper(fake_data: any): HelperOptions {
+  let fail_callback = () => {
+    throw new Error("spoofHelper is not sufficient here.");
```
Eranziel, line 399: You dabble in black magics, sir. 😆

PR #590, opened 2023-11-09
convert FullRepair from Macro to FlowState

Kuenaimaku:
Starting to help with converting macros over to flows with some low-hanging fruit.


src/module/actor/lancer-actor-sheet.ts
```diff
@@ -190,7 +190,7 @@ export class LancerActorSheet<T extends LancerActorType> extends ActorSheet<
       const flowArgs = flowElement.dataset.flowArgs;
       switch (flowType) {
         case "FullRepair":
-          break;
+          this.actor.beginFullRepairFlow(flowArgs);
```
Eranziel, line 193: Needs to keep the `break;` on a following line, otherwise it will also execute the "Stabilize" case.

src/module/flows/full-repair.ts
```diff
@@ -1,27 +1,49 @@
...
+    // Initialize data if not provided
+    const initialData: LancerFlowState.TextRollData = {
+      title: data?.title || "",
+      description: data?.description || "",
+      tags: data?.tags || [],
+    };
+
+    super("FullRepairFlow", uuid, initialData);
+    this.steps.set("displayFullRepairDialog", displayFullRepairDialog);
+    this.steps.set("executeFullRepair", executeFullRepair);
```
Eranziel, line 23: Note that I made a couple changes that will need to be incorporated here. 
- The base `Flow` constructor no longer has the first argument for the flow name. The flow name is now defined at the class level.
- Steps are set at the class level as an array of strings, which are fetched from the step registry at runtime.
- Each flow source file has a function which registers its steps in the flow registry. Those functions are imported and called in `lancer.ts`.

You can see examples of this in `flows/attack.ts`, `flows.activation.ts`, etc... on the latest 2.0-rc commits. Should be fairly straightforward, mostly just copy-pasting things to a different place.

Eranziel (changes requested):
Cool, thanks! Two things I noticed, one is because I changed things after you started working on this (sorry about that).

src/module/flows/full-repair.ts
Kuenaimaku: Latest commit should fix all of these issues!

src/module/actor/lancer-actor-sheet.ts
Kuenaimaku: good catch - updated in latest commit!

Eranziel (approved):
👍

PR #596, opened 2024-02-08
Update system file to allow system to be used in v11

Truuun:
I tested this system in  foundry version 11.315 and didn't see any immediate issues. I updated the system file so that it can now show up in the systems list in v11 too.


Truuun:
Alright, clear, thanks for the info, closing!

PR #598, opened 2024-03-11
Save Target and Sensors for Pilots

Davis-K-Johnson:
Added icons and values for the save target and sensors stats for the pilot sheet.

### DIFF
--- a/src/module/actor/lancer-actor.ts
+++ b/src/module/actor/lancer-actor.ts
@@ -249,6 +249,8 @@ export class LancerActor extends Actor {
       this.system.grit = Math.ceil(this.system.level / 2);
       this.system.hp.max = 6 + this.system.grit;
       this.system.bond = (this.items.find(i => i.is_bond()) ?? null) as unknown as LancerBOND | null;
+      this.system.sensor_range = 5;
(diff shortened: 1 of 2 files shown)


Eranziel (approved):
Looks good, thanks! I'll hopefully give it a test tonight, then merge if I don't have any other feedback.

src/module/actor/lancer-actor.ts
```diff
@@ -249,6 +249,8 @@ export class LancerActor extends Actor {
       this.system.grit = Math.ceil(this.system.level / 2);
       this.system.hp.max = 6 + this.system.grit;
```
msprijatelj, line 250: So unrelated to the PR, but @Eranziel did you know that there’s a [rules.json](https://github.com/massif-press/lancer-data/blob/master/lib/rules.json#L6) file in lancer-data that has some of these baselines already defined? Probably unlikely to change in the future but it might be helpful to avoid “magic numbers” in hardcoding?

Eranziel: I think I did know about that, but had forgotten about it! Thanks for the reminder. I'll see about integrating those where we can.

PR #599, opened 2024-03-12
Overheat Flow

Davis-K-Johnson:
Implemented the automation for overheating to work with flows, in line with how the structure one works. I also renamed the flow file to overheat instead of stress since that helped with clarity while coding, but if that's not what you want I can change it back.


src/module/flows/overheat.ts
```diff
@@ -0,0 +1,340 @@
...
+          return "Roll an ENGINEERING check. On a success, your mech is @Compendium[world.status.Exposed]. On a failure, it suffers a reactor meltdown after 1d6 of your turns (rolled by the GM). A reactor meltdown can be prevented by retrying the ENGINEERING check as a free action..";
+        case 1:
+          return "Your mech suffers a reactor meltdown at the end of your next turn.";
+        default:
+          return "Your mech becomes @Compendium[world.status.Exposed]";
+      }
+    case 2:
+    case 3:
+    case 4:
+      return "The power plant becomes unstable, beginning to eject jets of plasma. Your mech becomes @Compendium[world.status.Exposed.";
```
Eranziel, line 126: Missing the `]` after Exposed.

public/templates/chat/overheat-card.hbs
```diff
@@ -1,37 +1,39 @@
-<div class="card clipped-bot" style="margin: 0px;">
+<div class="card clipped-top" style="margin: 0px;">
   <div class="lancer-header lancer-primary struct-stress clipped-top">
-    {{#if rerollMacroData}}<span class="spacer"></span>{{/if}}
+    <i class="cci cci-stress i--m"></i>
+    {{#if reroll_data}}<span class="spacer"></span>{{/if}}
     <div class="flexcol">
-      <span>{{localize "lancer.chat-card.title.overheating"}}</span>
-      <span style="align-self: center;">{{localize "lancer.chat-card.remaining" total=val max=max}}</span>
+      <span>{{localize "lancer.chat-card.title.stress"}}</span>
```
Eranziel, line 6: Localization key in `en.json` for this is `lancer.chat-card.title.overheating`.

src/module/flows/overheat.ts
```diff
@@ -0,0 +1,340 @@
...
+  }
+
+  // After the hud is closed, update the actor's heat and stress.
+  if (!state.data?.reroll_data) {
+    let heat = actor.system.heat;
+    let stress = actor.system.stress;
+    if (heat.value > actor.system.heat.max && stress.value > 0) {
+      await actor.update({
+        "system.stress": stress.value - 1,
+        "system.heat": heat.value - actor.system.heat.max,
```
Eranziel, line 87: There's a special case for single-structure NPCs that currently isn't supported, but I think could be better supported fairly easily. When they go over heat cap they don't lose structure, they instead become exposed until they clear their heat.

We should copy the core book's wording for that case and display it in the card. Do you think it makes sense to add that to `overheatTableTitles` and `overheatTableDescriptions`? Or should we treat it as a truly special case and have separate variables that we use?

Eranziel (changes requested):
Nice work! Works really well, and that's a good chunk of work done to get the flows refactor done. I'm good with the file name change, that makes sense.

I noticed a few things in the code and while testing that can be fixed up, see comments below.

PR #600, opened 2024-03-14
Compendium Folders

BoltsJ:
- [x] Create folders for documents when importing Core Data and LCPs
- [x] Place imported Items into folders on import/update
- [ ] Use a smaller number of compendiums with more item types in each.
- [ ] Ensure that pilot importing and NPC creation correctly finds the needed items in the new compendium locations
- [ ] Migrate pre-2.0 compendiums to the new structure. 
- [ ] (Optional) Pretty banners (Zenn is working on these in the league discord)


BoltsJ:
image

Issue #582, opened 2023-09-01
[BUG] Lancer AoE Templates don't target on Gridless


msprijatelj:
**Describe the bug**
When on a Gridless Scene, the AoE Template macros included with Lancer do not automatically target tokens.

**To Reproduce**
1. Create a new, gridless scene.
2. Drag tokens onto the scene.
3. Execute any Lancer AoE Template macro, completely covering at least one token.
4. Observe that the token is not targeted.

**Expected behavior**
The Lancer AoE Template Macros target tokens regardless of whether a scene is Gridless or not.

**Platform (please complete the following information):**
 - OS: Windows 10
 - Browser Firefox
 - Foundry version 10.303
 - Lancer system version 1.6.1

**Additional Context**
The [Smart Target](https://github.com/theripper93/Smart-Target) module seems to have no issue on Gridless scenes; maybe some code could be lifted from there, along with any code that could help with #581 ?

**Discord contact info**
_valkyrion

PR #607, opened 2024-03-22
Gridless Template Targeting

BoltsJ:
This adds proper template targeting on gridless scenes. Depends on bug fixes from #606 

Fixes: #582


Eranziel:
LGTM!

PR #617, opened 2024-03-25
Configure trackableAttributes

BoltsJ:
Configures the trackableAttributes config value so that the correct resources show for each actor type in the token config. Also adds overshield.value as a value in addition to the bar version.


whitespine:
Looks good to me

PR #620, opened 2024-04-01
Add integration for Carousel Combat Tracker

BoltsJ:
Fixes up some combat details and adds a hook and an api for integration with the Carousel Combat Tracker module (combat-tracker-dock). For the integration, on the system side, the initiative value is initialized to 0 for all combatants if null so that the dock removes the roll initiative button (so we don't have to).

On the module side, we provide a function that gets a short description for deployables and npcs and a function to supply a list of buttons, which we use to show remaining activations and a deactivate button for the current combatant.

### DIFF
--- a/src/lancer.ts
+++ b/src/lancer.ts
@@ -661,7 +661,20 @@ Hooks.once("init", async function () {
   //   }
   // });
 
+  // Combat tracker HUD modules integration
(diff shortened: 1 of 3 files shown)


BoltsJ:
Also, this make previousTurn refund an activation and set the turn to null.

Issue #394, opened 2022-01-26
[FEATURE] NPC Traits Made Indestructible


msprijatelj:
**Is your feature request related to a problem? Please describe.**
This is relatively minor, but by the rules of Lancer's NPCs in the Core Book on Pg. 282: "Because Traits aren’t systems, they can’t be disabled or destroyed by damage." However, NPC Actor traits can be marked as destroyed on FoundryVTT.

**Describe the solution you'd like**
Disable the "Toggle Destroyed" option for NPC traits; toggle all NPC traits back to "intact" if they were previously marked as destroyed to prevent locking the status to "Destroyed".  This may make it easier to setup automation for weapon/system destruction (though there may be other ways to check for the "Trait" type).

**Describe alternatives you've considered**
Let it be; leaves chance for user confusion, but also for flexibility (e.g. if the user *wants* to mark the trait as destroyed or disabled, they are currently able to do so).

**Additional context**
Honestly, this comes down to how hard and fast the team wants Lancer to stick to the core rules of the game.  If the team wants to take a looser approach to the rules, feel free to mark this as `wontfix`.

**Discord contact info**
Valkyrion#6196

mason-bially:
I think this is an important note to have for if/when we *do* implement system/weapon destruction automation (NPC or otherwise). Though I suspect the existence of the destroyed toggle is fine and should probably be won't fix, since it's not going to simplify the underlying code.

PR #624, opened 2024-04-16
2.0 - Don't allow NPC Traits to be marked as destroyed

codyhod:
Resolves https://github.com/Eranziel/foundryvtt-lancer/issues/394

As per discussion there, this may not be something you intend to 'fix':

> I think this is an important note to have for if/when we do implement system/weapon destruction automation (NPC or otherwise). Though I suspect the existence of the destroyed toggle is fine and should probably be won't fix, since it's not going to simplify the underlying code.

However, I believe the underlying bug is still valid, in that NPC traits *are* indestructible, and to assist new players & DMs the Foundry System should mimic the expected behaviour from the System rules.

Regardless, this was super easy to do - so if you choose not to merge it, no worries.

---

Import the `NpcFeatureType` Enum, explicitly ignore NPC features with `item.system.type` "Trait" when returning the Context Menu.

### DIFF
--- a/src/module/helpers/item.ts
+++ b/src/module/helpers/item.ts
@@ -33,6 +33,7 @@ import {
   DamageType,
   EntryType,
   FittingSize,
+  NpcFeatureType,
   RangeType,
   ReserveType,
   SystemType,
@@ -1524,7 +1525,9 @@ function _handleContextMenus(
       return (
         !view_only &&
         item instanceof LancerItem &&
-        (item.is_mech_system() || item.is_mech_weapon() || item.is_npc_feature()) &&
+        (item.is_mech_system() ||
+          item.is_mech_weapon() ||
+          (item.is_npc_feature() && item.system.type !== NpcFeatureType.Trait)) &&
         !item.system.destroyed
       );
     },


Eranziel:
I am a bit torn on this one, because I think it's sometimes useful to be able to mark a trait "destroyed" for other bookkeeping purposes. That said, should probably just merge it. It would help avoid confusion.

msprijatelj:
Eh, my 2 copper these days is to keep the current functionality. More flexible that way.

codyhod:
Perhaps a more suitable solution is to conditionally change the wording of the context menu to avoid confusion?

AFAIK there isn't automation tied to destroyed features right now, so the issue comes from how the wording collides with the game rules. 

Off the top of my head, something like "Hide" that accomplishes the same function as "Mark as destroyed" currently so you have the flexibility to use it for bookkeeping as mentioned, while also not using any words a new player might expect to have meaning (Destroyed, disabled). This also means that just by clicking around on the context menus, any new players can tell that there is definitely a difference :)

Eranziel:
I do like the idea of keeping the functionality but just changing the wording. "Disabled" might be good, actually? "Destroyed" has a mechanical meaning in Lancer, a "disabled" system doesn't really (unless I'm forgetting something).

msprijatelj:
Last I checked “disabled” is unused, yes.

Eranziel:
@codyhod I didn't mean to close this PR, sorry. It should target `master` now, if you want to re-create it.

codyhod:
No worries, was thinking about how to make it work without writing bad code and got distracted with other projects.

The issue is still open, I'll come back to it at some point :)

Issue #424, opened 2022-05-01
[BUG]Removing a Class or Template from an NPC sheet no longer removes applicable features


Z3nner:
Removing a Class or Template from a character sheet no longer removes applicable features.

**To Reproduce**
1. Create a new NPC or open an existing one.
2. Add a class or template to the character sheet Observe the base features being populated
3. Remove the class or template.  Observe that the previously added features are not removed.

**Expected behavior**
I believe in the past these features were automatically removed with the class/template.

Davis-K-Johnson:
![Image](https://github.com/Eranziel/foundryvtt-lancer/assets/89803504/450be231-5628-4f0e-9073-6ce21da8e2e2)

I used the steps to recreate the problem and it appears to still be working like that. All of the mech stats reset to 0 as expected but systems are still left over.
OS: Windows 10
Browser: Chrome
Foundry version: 11.315
Lancer system version: 2.0.0

Eranziel:
Fixed in #625.

PR #625, opened 2024-04-17
2.0 - Remove linked class/template features on item deletion

codyhod:
Resolves https://github.com/Eranziel/foundryvtt-lancer/issues/424

Rework unused function `handleClassDelete` to take an item (the template or class) and an npc instead of using Context.

Conditonally call this function when we click "Delete Document" in a context menu.


Eranziel (commented):
I like the changes, but I've been meaning to move the class change/feature deletion logic to LancerActor instead of the NPC sheet logic. Watch this space. ;)

PR #630, opened 2024-05-31
Release automation

Eranziel:
Creating a PR from Zenn's branch to get a better view of the changes.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,7 +3,7 @@ name: Release Creation
 on:
   push:
     tags:
-      - 'release-*'
+      - 'v*'
 
 jobs:
   build:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,110 @@
+name: Release Creation
+
+on:
+  push:
+    tags:
+      - 'release-*'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+
+    # Load the system.json manifest into memory
+    - name: Load system manifest
+      id: manifest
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,110 @@
...
+    - name: Build Packs
+      run: gulp build_packs
+
+    # Package the built files
+    - name: Package Files
+      run: gulp package
+
+    # Run git operations
+    - name: Git Operations
+      run: gulp git
```
Eranziel, line 76: I think I'll want to change this a bit for automation. `gulp manifest` and `gulp git` should get run locally and then that commit tagged to trigger the release. Probably abort the workflow if the tag and version in `system.json` do not match, as a safeguard against building malformed packages.

.github/workflows/release.yml
```diff
@@ -0,0 +1,110 @@
...
+        RELEASE_DOWNLOAD: ${{ env.RELEASE_DOWNLOAD_URL }}
+        PACKAGE_VERSION: ${{ steps.manifest.outputs.version }}
+        PACKAGE_DOWNLOAD: ${{ steps.manifest.outputs.download }}
+      run: |
+        if [[ ! $TAG_NAME == release-$PACKAGE_VERSION ]]; then
+          echo "The system.json version does not match tag name."
+          echo "system.json: $PACKAGE_VERSION"
+          echo "tag name: $TAG_NAME"
+          echo "Please fix this and push the tag again."
+          exit 1
```
Eranziel, line 45: Ah, this should already bail if `system.json` and the tag name don't match, great!

Eranziel (commented):
Recording some thoughts so I don't forget!

Eranziel:
Merged to master instead, this PR is no longer needed.

PR #632, opened 2024-06-01
Remove more duplicates

C-S-McFarland:
More status effect duplicates. Thanks to @msprijatelj for pointing these out.


C-S-McFarland:
I pulled the most recent 2.0-rc and I think this is already fixed. Not sure how GitHub let me make this.

PR #652, opened 2024-06-04
[RFC] Add pre- and post-flow hook calls

cirrahn:
These hooks provide a familiar Foundry API for interacting with the new flow system, allowing module developers to trigger simple side-effects on flow step changes.

The "pre" hook additionally allows the flow to be cancelled, mimicking the functionality of "pre" hooks in Foundry (for example, "preUpdateActor").

image

See also: https://github.com/Z3nner/lancer-weapon-fx/pull/36


BoltsJ:
Superseded by #656 , but it did help me think about the design for the api.

PR #671, opened 2024-06-06
Update statusEffects to use name

C-S-McFarland:
Some of the additional status effects were using label instead of name.


BoltsJ:
Are these localizable? If they are, may as well add localization for them while touching this part of the code.

C-S-McFarland:
Thank you @BoltsJ for the help.

Eranziel (approved):
LGTM!

PR #716, opened 2024-07-04
Add more white icons

Z3nner:
For all icons currently available in black, add a white version in icons\white.
Also, for your consideration, added corepower.svg in both black and white.

### DIFF
--- a/public/assets/icons/white/shield_outline.svg
+++ b/public/assets/icons/white/shield_outline.svg
@@ -1,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   version="1.1"
+   width="24"
+   height="24"
+   viewBox="0 0 24 24"
+   id="svg1"
+   sodipodi:docname="shield_outline.svg"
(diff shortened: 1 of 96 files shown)


public/assets/icons/white/shield_outline.svg
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   version="1.1"
+   width="24"
+   height="24"
```
BoltsJ, line 5: They are svgs so its not the end of the world, but it would be better imo to have this the same 512 x 512 size as the others.

public/assets/icons/white/tag.svg
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   width="24"
+   height="24"
```
BoltsJ, line 4: Size again

BoltsJ (changes requested):
Two of the icons have a smaller size. Other than that this looks good to me

Issue #650, opened 2024-06-04
[FEATURE] Foundry V12 Support


BoltsJ:
Main list of things for v12 support

[V12 Breaking Changes](https://github.com/orgs/foundryvtt/projects/42/views/9)
 
- [x] Remove deprecations from v10 - foundryvtt/foundryvtt#10164
- [x] Use TypeDataModel instead of DataModel for document data - [System Data Model Documentation](https://foundryvtt.com/article/system-data-models/)
  - [ ] (optional) Move type-specific portions of prepareBaseData and prepareDerivedData to the data models
  - [ ] (Optional) Cleanly handle subtypes not provided by the system (`base` or module provided subtypes for example)
- [x] (Optional) Remove template.json and define document types in system.json - https://github.com/foundryvtt/foundryvtt/issues/10444 [Manifest defined types](https://foundryvtt.com/article/module-sub-types/)
- [x] Fix Measured templates - Canvas changes, foundryvtt/foundryvtt#10088
  - [x] Update Measured Templates to use grid based shapes foundryvtt/foundryvtt#10476
- [x] Update manifest grid configuration - foundryvtt/foundryvtt#9999
- [x] Remove square diagonal system setting
- [ ] (Optional) Migrate applications to app v2, possibly using a Svelte backend for rendering contents - foundryvtt/foundryvtt#5441
- [x] Ensure node version in workflows and guides is bumped to at least 18
- [ ]  (Optional) Custom data model for Combat, Combatant, ChatMessage and ActiveEffect - foundryvtt/foundryvtt#10185
- [x] Investigate whether the dynamic token ring system can be leveraged to create hex borders for tokens – It cannot
- [ ] (Optional) Add \@\Embed support for Documents, e.g. creating statblocks for npcs in journals. - foundryvtt/foundryvtt#10262
- [x] Use data fields for settings. See core settings for examples
- [x] Round borders on gridless
- [ ] Use max and min properties of ArrayField in favor of the custom BoundedArrayField

Bugs 🐛 :
- [x] Pilot imports don't properly copy the permission config from the pilot to the imported mechs and deployables.

Eranziel:
Thanks so much for this list, very helpful! All in all, doesn't look too bad.

BoltsJ:
It may or may not be a complete list. This is just based on memory (I have ADHD, so) and reading the breaking changes list.

C-S-McFarland:
[...]

PR #717, opened 2024-07-04
fix: Add grid config for v12 to system manifest

TravisWhitehead:
This addresses deprecation warnings appearing in Foundry v12 related to this item from:

https://github.com/Eranziel/foundryvtt-lancer/issues/650

> * [ ]  Update manifest grid configuration - [V12 System manifest schema changes foundryvtt/foundryvtt#9999](https://github.com/foundryvtt/foundryvtt/issues/9999)

I assume we want to keep the old `gridDistance` and `gridUnits` for now for Foundry v11 support?

## Testing

Verified that the deprecation warnings (for `gridDistance` and `gridUnits`) disappeared after adding this (in Foundry v12).

### DIFF
--- a/public/system.json
+++ b/public/system.json
@@ -53,6 +53,10 @@
     }
   ],
   "socket": true,
+  "grid": {
+    "distance": 1,
+    "units": ""
+  },
   "gridDistance": 1,
   "gridUnits": "",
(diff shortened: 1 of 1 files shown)


BoltsJ:
V12 development has not started yet.

PR #718, opened 2024-07-06
feat(macros): add optional actorPermLevel change to Scans

msprijatelj:
Add an optional parameter, `actorPermissionLevel`, to both Scan macros. When set to an integer greater than or equal to 0, it sets the Default Permission Level to `actorPermissionLevel`. Default value of `actorPermissionLevel` is `-1`.

### DIFF
--- a/src/packs/core_macros/Scan_3KPsYon0LoANWX54.yml
+++ b/src/packs/core_macros/Scan_3KPsYon0LoANWX54.yml
@@ -150,9 +150,9 @@ ownership:
   2a2IgZ2kiAqXSGqu: 3
 _stats:
   systemId: lancer
-  systemVersion: 2.0.0
-  coreVersion: "11.315"
+  systemVersion: 2.1.5
+  coreVersion: '11.315'
   createdTime: null
-  modifiedTime: 1717340495229
-  lastModifiedBy: RyAORVJNmdo1gq2Z
-_key: "!macros!3KPsYon0LoANWX54"
+  modifiedTime: 1720297619727
+  lastModifiedBy: VTqqjv3vCWDxBIi5
+_key: '!macros!3KPsYon0LoANWX54'
--- a/src/packs/core_macros/Scan_3KPsYon0LoANWX54.yml
+++ b/src/packs/core_macros/Scan_3KPsYon0LoANWX54.yml
@@ -6,7 +6,7 @@ img: systems/lancer/assets/icons/macro-icons/tech_quick.svg
 scope: global
 command: |-
   let targets = Array.from(game.user.targets);
-
+  const actorPermissionLevel = -1
   function sort_features(a, b) {
     return b.system.origin.base - a.system.origin.base;
   }
@@ -56,6 +56,11 @@ command: |-
 
   targets.forEach((target) => {
     let actor = target.actor;
+    if (actorPermissionLevel >= 0) {
+      const actorId = actor.system.parent._id
+      const parentActor = game.actors.get(actorId)
+      parentActor.update({ ownership: { default: actorPermissionLevel } });
+    }
     let items = actor.items;
     let hase_table_html = `
   <table>
--- a/src/packs/core_macros/Scan_3KPsYon0LoANWX54.yml
+++ b/src/packs/core_macros/Scan_3KPsYon0LoANWX54.yml
@@ -4,12 +4,21 @@ type: script
 author: 2a2IgZ2kiAqXSGqu
 img: systems/lancer/assets/icons/macro-icons/tech_quick.svg
 scope: global
-command: |-
+command: >-
   let targets = Array.from(game.user.targets);
-  const actorPermissionLevel = -1;
+
+
+  //actorPermissionLevel - This sets the permission level for the parent actor
+  of a token (from CONST.DOCUMENT_PERMISSION_LEVELS, use NONE, LIMITED,
+  OBSERVER, or OWNER). If set to `null`, no change is made.
+
+  const actorPermissionLevel = null;
+
+
   function sort_features(a, b) {
     return b.system.origin.base - a.system.origin.base;
   }
+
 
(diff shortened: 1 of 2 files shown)


src/packs/core_macros/Scan__Journal__z0pdvwE0FitMzwRJ.yml
```diff
@@ -5,106 +5,61 @@ author: jE81C9hyfb9evTbT
...
-
   // ======================================================
-
   //Variables - Change these to control the macro
-
   // ======================================================
 
-
   // outputToChat - if set to true, also send the scan results to chat.
-
```
BoltsJ, line 27: It looks like you directly edited the yml file, so this won't actually run because all these lines get smashed into a single line. You need to edit the macro in foundry and then unpack the db with `npx gulp export_packs`

image

BoltsJ (changes requested):

BoltsJ:
Also, semicolins

src/packs/core_macros/Scan_3KPsYon0LoANWX54.yml
```diff
@@ -6,18 +6,16 @@ img: systems/lancer/assets/icons/macro-icons/tech_quick.svg
 scope: global
 command: |-
   let targets = Array.from(game.user.targets);
-
+  const actorPermissionLevel = -1;
```
BoltsJ, line 9: You may want to document this and possibly use CONST.DOCUMENT_PERMISSION_LEVELS to make it more clear.

src/packs/core_macros/Scan_3KPsYon0LoANWX54.yml
```diff
@@ -54,8 +52,13 @@ command: |-
     return sc_templates;
   }
 
-  targets.forEach((target) => {
+  targets.forEach(target => {
     let actor = target.actor;
+    if (actorPermissionLevel >= 0) {
+      const actorId = actor.system.parent._id;
+      const parentActor = game.actors.get(actorId);
+      parentActor.update({ ownership: { default: actorPermissionLevel } });
```
BoltsJ, line 60: Is it possible to modify only the permission on the token actor? That way unlinked tokens don't get their status permanently revealed. It's not something I've really played with, but I think that would probably be less annoying for enemies that get reused.

src/packs/core_macros/Scan__Journal__z0pdvwE0FitMzwRJ.yml
```diff
@@ -60,6 +60,12 @@ command: >-
 
   const updateExisting = true;
 
+  //actorPermissionLevel - This sets the permission level for the parent actor
+  of a token. This must be an integer between -1 and 3 where -1 is "No change",
+  0 is "None", 1 is "Limited", 2 is "Observer", and 3 is "Owner"
+
+  const actorPermissionLevel = -1;
```
BoltsJ, line 67: May want to use the CONST values for readability

BoltsJ (commented):
Just some thoughts. Overall looks fine.

src/packs/core_macros/Scan_3KPsYon0LoANWX54.yml
BoltsJ: Also, perhaps it would be better to use `undefined` or `null` explicitly to sentinel the feature as turned off?

src/packs/core_macros/Scan_3KPsYon0LoANWX54.yml
msprijatelj: I can, but that's not the use case I'm looking for with this feature. My thought on this is:
1. If the players scan a token, they know all of that token's stats, yes. However, if this token has an unlinked parent actor, they basically already know the parent's stats, too.
2. If they already know the parent's stats, they can track any children's HP/heat/stat changes by just keeping track of damage dealt/healed.
3. Therefore, by granting Observer permissions on the parent, this cuts out the busywork of taking notes and "doing the math".

My take is that if a GM who wants to enable this feature wants to truly obscure stats of different tokens, they'd use a different parent token with different stats.

src/packs/core_macros/Scan_3KPsYon0LoANWX54.yml
msprijatelj: I'll document it to make it more clear, sure. `null` can work fine. Values should be something like the following, right?
```
CONST.DOCUMENT_PERMISSION_LEVELS.NONE
CONST.DOCUMENT_PERMISSION_LEVELS.LIMITED
CONST.DOCUMENT_PERMISSION_LEVELS.OBSERVER
CONST.DOCUMENT_PERMISSION_LEVELS.OWNER
```

src/packs/core_macros/Scan__Journal__z0pdvwE0FitMzwRJ.yml
msprijatelj: Should be addressed now! Resolving

src/packs/core_macros/Scan_3KPsYon0LoANWX54.yml
msprijatelj: Documented with a comment and used the suggested constants!

src/packs/core_macros/Scan_3KPsYon0LoANWX54.yml
msprijatelj: I could understand maybe removing this option from the Default scan macro, to keep that macro more laser-focused on providing a chat message Scan. The Journal Scan macro, however, is for folks who want to build up a "Pokedex" of NPCs, so those scans will provide clear statblocks that players can then use to track changes to stats like HP and Heat. In that use case, optionally providing ownership permissions at the parent actor level makes the most sense to me.

BoltsJ (approved):

PR #732, opened 2024-07-18
Fix tech attack not consuming lock on

C-S-McFarland:
When the tech attack card is made, I think it has the wrong targets returned.

### DIFF
--- a/src/module/flows/system.ts
+++ b/src/module/flows/system.ts
@@ -82,7 +82,7 @@ async function printSystemCard(
     // attackData: {
     //   origin: state.actor.id,
     //   targets: state.data.attack_rolls.targeted.map(t => {
-    //     return { id: t.target.id, lockOnConsumed: !!t.usedLockOn };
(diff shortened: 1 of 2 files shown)


BoltsJ (approved):
Looks like there's another usage of `lockOnConsumed` that's currently commented out in the system flow. Might be worth updating that one too since it's in a block labeled TODO
```
b@toolbox ~/D/f/foundryvtt-lancer (foundry-v12)> rg lockOnConsumed
src/module/flows/system.ts
85:    //     return { id: t.target.id, lockOnConsumed: !!t.usedLockOn };

src/module/flows/tech.ts
186:        return { id: t.target.id, lockOnConsumed: !!t.usedLockOn };
```
Other than that, looks good to me

Eranziel:
Yup, that's the fix! I actually ran across this the other day while working on damage application and it didn't occur to me to cherry pick the fix over to master.

C-S-McFarland:
@BoltsJ I updated the commented out code as well. :+1:

BoltsJ (approved):

Eranziel (approved):

PR #736, opened 2024-07-23
[WIP] adding buttons to apply damage from attacks to actors

lmilekic:
I was looking at [this](https://github.com/Eranziel/foundryvtt-lancer/issues/189) a while ago and thought it would be really nice to have some buttons for 1/2, full, 2x damage like pathfinder has. This is a very barebones PR to get feedback on that thinking.

Putting this PR out there to bother @Eranziel on how to actually:
1. Grab actors targeted by the attack
2. Do damage to those actors

Right now it looks pretty ugly, but I figure we can make the buttons smaller and prettier once the actual logic is working:
image


Eranziel:
Thanks for the PR!

Unfortunately, I've been working on this lately (in the `damage-application` branch, if you want to take a look), which looks like it will supersede your work here.

lmilekic:
Glad to hear it! I'll take a look at your branch!

PR #739, opened 2024-07-27
Allow previewing LoS in attack HUD

BoltsJ:
When the attack hud is active and Terrain Height Tools enabled in the
world, hovering over the targets or the cover indicator in the hud will
show the THT LoS preview between the attacker and the hovered target.


BoltsJ:
[Screencast from 2024-07-27 17-34-23.webm](https://github.com/user-attachments/assets/577134d9-7df0-4ca7-b429-0bdde32beb3d)

Issue #740, opened 2024-07-29
[BUG] - Attack modal checkboxes


tradiuz:
**Describe the bug**
The attack modal diaglog has the X marks for the checkbox mis-aligned.

**To Reproduce**
1. Select a target token
2. Use a weapon or tech attack

**Expected behavior**
Xs should be centered in the box.

**Screenshots**
![image](https://github.com/user-attachments/assets/72d7ee64-c135-45bc-9d78-34c20279563d)

**Platform (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome and Firefopx
 - Foundry version: 10.315 (installed as container)
 - Lancer system version: 2.2.0

**Additional context**
Tested with 0 modules, direct to container port to rule out modules and proxy.

**Discord contact info**
tradiuz

Eranziel:
Closed by #742.

PR #742, opened 2024-07-29
Use native checkboxes and styling

BoltsJ:
Fixes wild differences in some browsers, I haven't tested it on windows, but on Firefox and Chromium on Linux the differences are almost imperceptible. I have set the parameters to be optimal for Chromium.
Chromium Linux:
Screenshot from 2024-07-29 14-19-18
Firefox Linux:
Screenshot from 2024-07-29 14-20-04


Eranziel:
Closes #740.

PR #756, opened 2024-08-23
build: Update aws-amplify from 4.3.36 to 5.3.21

lazulit3:
Updates `aws-amplify` dependency from `4.3.36` to `5.3.21`

This is not the latest version, but is low hanging fruit compare to major version 6.

## Motivation

This is an attempt at helping with items mentioned at https://github.com/Eranziel/foundryvtt-lancer/pull/706

> - Maintenance: Keeping the aws dependencies updated represents a non-trivial amount of work.
> - Security: There are several current security notices for this repo due to aws-amplify and its dependencies.

## Results

This reduces the number of vulnerabilities detected by `npm audit`

Before (on `master` @ 67da63b4f03288558cadd3d96df9f646c543d0d9):

```
57 vulnerabilities (49 moderate, 8 high)
```

Afterwards (on this PR branch):
```
40 vulnerabilities (33 moderate, 7 high)
```

Additionally, `aws-amplify` is no longer directly responsible for any reported vulnerabilities at this time:
```sh
npm install && npm audit | grep aws-amplify
```

## Notes

[List of breaking changes from `aws-amplify` v4 -> v5](https://github.com/aws-amplify/amplify-js?tab=readme-ov-file#amplify-5xx-has-breaking-changes-please-see-the-breaking-changes-below)

The only breaking change found/addressed here is a trivial adjustment to `Storage.list` return type.

### Thoughts on v6

I took a stab at breaking changes for v6 as well, and I agree it's a non-trivial amount of work.
If desired I could continue that WIP-- but it looks like COMPCON is still using a Gen 1 Amplify setup which makes the usage a little weirder. (I am curious if COMPCON v3 will be using Amplify Gen 2, but I don't know where to find COMPCON v3's source code.)

Please let me know if help moving to aws-amplify v6 is desired.

## Testing

I tested the following:

- Building/installing system on Foundry v11.315
- Logged into COMP/CON
- Visited sync menu on Pilot Actor Sheet
- Downloaded various pilots from my COMP/CON cloud data
- Verified my pilots/mechs appeard in Actors


BoltsJ:
Part of the problem with amplify 5 is that it depends on a vulnerable version of axios, but the axios dependency is entirely removed in amplify 6.

lazulit3:
>Part of the problem with amplify 5 is that it depends on a vulnerable version of axios

I think this may no longer be the case, or I am reading output wrong ;)

Also I think the vuln count I put in PR desc is wrong, maybe because my `node_modules` was dirty from other branches.

With a clean install of deps on this branch, I see:

```
9 vulnerabilities (2 moderate, 7 high)
```

`micromatch` is only dep w/ vuln I see right now that's related to `aws-amplify` v5-- and that has a fixed version available.

Am not gonna push more version bumps w/o testing so will leave this as-is-- but I think the nested dependencies may have been patched up since you last looked? 🎉

lazulit3:
> Part of the problem with amplify 5 is that it depends on a vulnerable version of axios, but the axios dependency is entirely removed in amplify 6.

If what you mean by this is that we should get to aws-amplify v6-- noted, I will take a stab at it.

lazulit3:
@BoltsJ Here's what amplify v6 would look like-- but has issues with no way to set `Cache-Control` header:
https://github.com/Eranziel/foundryvtt-lancer/pull/759

IMO we close the v6 PR and roll with this v5 PR unless we see a path to dealing with the caching issue discussed in linked PR.

BoltsJ (approved):
Now that there's a version of axios that's not vulnerable this is fine. I merged in master and ran an npm audit.

Issue #751, opened 2024-08-14
[BUG] Action Tracker Bar not displaying and Action buttons in sheet not working


Keyaira14:
**Describe the bug**
While in an encounter, the action bar which should appear at the bottom showing actions available to players for their turn, does not appear. In addition, the buttons under the Actions header of the <MECH//STATS> page of the character sheet do not function, doing nothing when pressed, and the movement button showing 0 movement left regardless of amount of movement used.

**To Reproduce**
1. Create Mech
2. Add Mech to map
3. Toggle Combat State
4. Start Encounter
5. Take turn
6. See error

**Expected behavior**
When taking turn, action bar should display showing what actions are left available for the turn. action buttons in sheet should function, as well as display movement left over.

**Screenshots**
![image](https://github.com/user-attachments/assets/fa3b4dc1-faba-4d1e-a06a-6ae4edcda1f5)
![image](https://github.com/user-attachments/assets/63758ed0-a223-4808-b854-ebecf8f2117c)

**Platform (please complete the following information):**
 - OS: Windows
 - Browser: Opera, Firefox and native Foundry app
 - Foundry version: 11 build 315
 - Lancer system version: 2.2.4

**Additional context**
This happens regardless of active Modules, and I admittedly have never seen the Action Tracker bar apart from screenshots, and do not know what is actually supposed to happen when pressing any of the buttons under the Action header of the character sheet (presumably show standard actions available to your character under that category) as this is my first time setting up a Lancer campaign using Foundry VTT.

A third party player also connected to the game, and encountered the same issues.

**Discord contact info**
keyaira

Eranziel:
Huh, I thought there was already an issue for this, but apparently not! Thanks for opening one.

Eranziel:
Closed by #757.

PR #757, opened 2024-08-26
fix: restore (broken) Action Tracker functionality

lazulit3:
This PR fixes functionality related to the Action tracking:
- https://github.com/Eranziel/foundryvtt-lancer/issues/751

I tried to understand the prior behavior and fix what bugs I could find, but am unsure if I got everything.

## Fixes & Improvements

- Action tracker now appears after selecting a Mech token that is engaged in combat
- Action tracker shows colored buttons when actions are available, and gray buttons when unavailable
- Actor's action tracker data now gets updated on turn start and end
- Action tracker's reset button now respects the "Allow players" setting (if disabled they cannot reset)
  - The reset button is no longer rendered if players are not allowed to use it
- Action tracker is displayed when NPC tokens are selected as well
- End of turn message (from "Print Message" Action Tracker setting) differentiates style for spent (colored) vs unspent (gray) actions

## Questions / Discussion

For the `Print Messages` setting in Action Tracker Settings, what is the desired behavior for the end-of-turn action output?
- Should the action badges for unused actions be colored and spent actions grayed out (to draw attention to unused action economy?)
- Or should end-of-turn message provide history about what actions a player took, and color the spent actions / gray out the unused?

Or neither? I see that the code intended to style them differently based on which actions were spent, but that style is gone/broken, so am guessing here.

## Testing

Behavior that I've verified (related to fixes):

- Action tracker appears when a mech or NPC is selected/controlled and is in a (started) combat encounter
  - Actions are colored when available, gray when unavailable
- Actions are refreshed when an actor starts their turn
- Actions are cleared (and reaction refreshed) when an actor's turn ends
- Actions buttons at bottom of `<<MECH//STATS>>` on Sheet update action tracker state
- If action tracker settings has `Allow players` disabled:
  - the reset button is not rendered
  - user cannot use the reset button (even if it was rendered)
- If action tracker settings has `Print Messages` enabled:
  - Message with action info appears in chat at end of turn
  - Action badge style reflects which actions were taken

### DIFF
--- a/public/templates/window/action_manager.hbs
+++ b/public/templates/window/action_manager.hbs
@@ -8,23 +8,23 @@
   <div style="display:inline-flex;">
     <i id="action-manager-drag" class="mdi mdi-drag"></i>
     <div class="action-bar">
-      <a class="action {{#if actions.protocol}}activation-protocol{{/if}}" data-action="protocol">
+      <a class="action {{#if actions.protocol}}lancer-protocol{{/if}}" data-action="protocol">
         <i class="cci cci-protocol theme--dark white--text"></i>
       </a>
       <hr class="v-divide" role="separator" aria-orientation="undefined" />
-      <a class="action {{#if actions.move}}activation-move{{/if}}" data-action="move">
+      <a class="action {{#if actions.move}}lancer-move{{/if}}" data-action="move">
         <i class="mdi mdi-arrow-right-bold-hexagon-outline theme--dark white--text"></i>
       </a>
-      <a class="action {{#if actions.full}}activation-full{{/if}}" data-action="full">
+      <a class="action {{#if actions.full}}lancer-full{{/if}}" data-action="full">
         <i class="mdi mdi-hexagon-slice-6 theme--dark white--text"></i>
       </a>
-      <a class="action {{#if actions.quick}}activation-quick{{/if}}" data-action="quick">
+      <a class="action {{#if actions.quick}}lancer-quick{{/if}}" data-action="quick">
         <i class="mdi mdi-hexagon-slice-3 theme--dark white--text"></i>
       </a>
-      <a class="action {{#if actions.reaction}}activation-reaction{{/if}}" data-action="reaction">
+      <a class="action {{#if actions.reaction}}lancer-reaction{{/if}}" data-action="reaction">
         <i class="cci cci-reaction theme--dark white--text"></i>
       </a>
-      {{!-- <a class="action {{#if actions.free}}activation-free{{/if}}" data-action="free">
+      {{!-- <a class="action {{#if actions.free}}lancer-free{{/if}}" data-action="free">
         <i class="cci cci-free-action theme--dark white--text"></i>
       </a> --}}
     </div>
--- a/public/templates/window/action_manager.hbs
+++ b/public/templates/window/action_manager.hbs
@@ -3,7 +3,9 @@
   style="width:{{position.width}}px; height:{{position.height}}px; top:{{position.top}}px; left:{{position.left}}px;">
   <div class="action-label">
     <div>{{name}}</div>
+    {{#if clickable}}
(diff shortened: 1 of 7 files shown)


lazulit3:
I think I've restored all of the prior functionality I could find-- please let me know if there's anything I missed.

src/lancer.scss
```diff
@@ -2432,8 +2433,9 @@ a.activation-chip > i.mdi {
 .track-message.lancer-action-grid {
   flex-wrap: wrap;
 }
-.track-message .lancer-action-badge {
-  margin: 4px;
+.lancer-action-badge {
+  background-color: var(--light-gray-color);
+  margin: 3px;
```
lazulit3, line 2438: The reason I removed `.track-message` here is to reduce specificity so that `button.lancer-#{$theme-type}` takes precedence for `background-color`. (So gray vs the action's color).

`.lancer-action-badge` is not used elsewhere, so removing `.track-message` here seems safe.

public/templates/window/action_manager.hbs
```diff
@@ -3,28 +3,30 @@
...
     <div>{{name}}</div>
+    {{#if clickable}}
     <div style="right: 0px;position: absolute;"><a id="action-manager-reset" class="mdi mdi-restore"></a></div>
+    {{/if}}
   </div>
   <div style="display:inline-flex;">
     <i id="action-manager-drag" class="mdi mdi-drag"></i>
     <div class="action-bar">
-      <a class="action {{#if actions.protocol}}activation-protocol{{/if}}" data-action="protocol">
+      <a class="action {{#if actions.protocol}}lancer-protocol{{/if}}" data-action="protocol">
```
lazulit3, line 13: I believe these classes like `activation-protocol`, `activation-full`, etc. (`activation-<action>`) are not defined, this is probably from old code.

May be worth grepping for other occurrences of these elsewhere in the code to clean up.

lazulit3 (commented):
Clarifying comments

lazulit3:
Oh I forgot to fix the active vs inactive styling for the Mech Sheet Action buttons-- should make that behavior consistent.

Will follow up later this week.

BoltsJ:
There's previous work on the `fix-action-bar` branch. Probably worth building off that rather than starting from scratch

lazulit3:
> There's previous work on the `fix-action-bar` branch. Probably worth building off that rather than starting from scratch

Thanks, I hadn't seen this. I've updated this branch with:
- Your WIP commit from `fix-action-bar` rebased on `master`
- My changes rebased on top of that

The only changes I have here that disagree w/ prior work is https://github.com/Eranziel/foundryvtt-lancer/pull/757/commits/dd3532d293787ddcf2ddc55af38e58139bb28729, my rationale for removing `this.updateActions(token.actor, _defaultActionData(token.actor))` is at https://github.com/Eranziel/foundryvtt-lancer/commit/21ebd6080f9cda943a0fc0c2992e7500401950d4#r145967069

lazulit3:
> Oh I forgot to fix the active vs inactive styling for the Mech Sheet Action buttons-- should make that behavior consistent.

Okay, got this done.

**I think this is ready for review.** ✨ 


### Style Examples

Fresh actions at turn start:

bar-fresh-actions

sheet-actions-fresh

Some actions spent:

bar-with-spent

sheet-actions-with-spent

Chat messages:

chat-turn-start-and-end

public/templates/window/action_manager.hbs
lazulit3: I removed the other occurrences of these when fixing the sheet action buttons.

Eranziel (approved):
Thanks for your work on this! And also apologies it's taken so long to get merged. It all seems to work properly from my testing!

I'm going to make some visual improvements to the action buttons on the mech sheet to make them align a bit more closely to the action tracker, but I'm happy to do that after merging.

PR #759, opened 2024-08-30
build!: Update aws-amplify from 4.3.36 to 6.5.2

lazulit3:
This is what it would look like to ugprade to `aws-amplify` v6-- but I do not think we want this.

I'm opening this for future/historical/archival reference purposes, but I think lack of `Cache-Control` config (discussed below) is a reason to close this and stick with https://github.com/Eranziel/foundryvtt-lancer/pull/756

**BREAKING CHANGE:** Cache-Control header no longer gets set on some storage requests, resulting in issues with latest version of pilot not being obtained from COMP/CON (in some cases).

## Blocker: Cache-Control

Amplify v6 deprecated the `cacheControl` option for Storage, so it's not possible (or not obvious to me) how to set `Cache-Control: no-cache` on storage requests.

The most that the [migration docs](https://docs.amplify.aws/gen1/javascript/build-a-backend/storage/storage-v5-to-v6-migration-guide/) (unhelpfully) have to say about this is:

> cacheControl: This option has been removed and now defers to the HTTP caching behavior configured by the runtime.

I can produce caching issues by updating a pilot on COMP/CON and trying to download again, and the updated pilot is not fetched due to disk cache.

Related issue @ amplify-js:

- https://github.com/aws-amplify/amplify-js/issues/13581


BoltsJ:
Closing in favor or #756

PR #780, opened 2024-11-13
Skip structure roll for NPCs with a single structure

C-S-McFarland:
I was asked to update Lancer QoL with this, but I think this is a rule that should probably be in the system.

The rule states:
> NPCs only have 1 Structure and are destroyed when they reach 0 HP. NPCs with additional Structure follow the standard rules for structure damage.

I check before the Structure roll HUD is displayed to just decrement the structure and return true to hit the rest of the flows.


C-S-McFarland:
Going to have to re-think this I think. Doesn't seem like a big enough deal.

Issue #587, opened 2023-10-19
[FEATURE] Default Scene Config


Eranziel:
Add a hook on scene creation that disables vision by default.

PR #783, opened 2024-11-14
Add hook pre scene creation to disable token vision and fog explorati…

sealtrowitz:
…on by default

Addresses #587


Eranziel (approved):
Ooh, shiny! Thanks!

PR #796, opened 2024-11-19
Fix dark theme setting legibility

### LINKED ISSUE #782: [BUG] Dark system UI themes have illegible text due to dark text color on dark background
**Describe the bug**
Dark system UI themes (GMS Dark, MSMC Solarized, HORUS Terminal [present but not as bad], Harrison Armory Ras Shamra, and FORECAST/GALSIM) have text that is too dark to be legible against the dark background in some setting windows.

**To Reproduce**
1. Set UI theme to MSMC Solarized
2. Open System Automation Settings
3. See error

**Expected behavior**
Text should be legible in all places on dark UI themes.

**Screenshots**
image

 - OS: Windows 11
 - Browser: Electron app
 - Foundry version: 11.315
 - Lancer system version: 2.4.0

**Additional context**
N/A

**Discord contact info**
skedoodlebug

sealtrowitz:
Fix #782. Temporary / imperfect accessibility solution until appv2 comes with v12.

### DIFF
--- a/src/lancer.scss
+++ b/src/lancer.scss
@@ -156,7 +156,7 @@ button {
   font-family: var(--main-font-family);
 }
 
-.window-app.lancer .window-content {
+.window-app.lancer .window-content .form-notes {
   background-color: var(--background-color);
   color: var(--dark-text);
 }
--- a/src/lancer.scss
+++ b/src/lancer.scss
@@ -156,7 +156,7 @@ button {
   font-family: var(--main-font-family);
 }
 
-.window-app.lancer .window-content .form-notes {
+.window-app.lancer .window-content {
   background-color: var(--background-color);
   color: var(--dark-text);
 }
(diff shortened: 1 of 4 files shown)


src/lancer.scss
```diff
@@ -156,7 +156,7 @@ button {
   font-family: var(--main-font-family);
 }
 
-.window-app.lancer .window-content {
+.window-app.lancer .window-content .form-notes {
```
Eranziel, line 159: This does narrow the style to one of our custom apps, but there's a bunch of other apps that need to get targeted too. I started building a list, but then it occurred to me - why don't we just remove the `lancer` class from the settings apps? `src/module/apps/action-tracker-settings.ts`, `.../automation-settings.ts`, and `.../status-icon-config.ts` should be all, and I think that's more logical than specifying another class for every `.lancer` app to apply the styling.

Eranziel (changes requested):

src/lancer.scss
sealtrowitz: Ah, I see what you're saying. That does make more sense,; I'll make the adjustments.

PR #808, opened 2024-12-28
Use downandout as the defeated status

BoltsJ:
I thought I set this in the v11 update, but I apparently dropped the
commit.


Eranziel (approved):
Feels weird to me because I'm used to the skull, but I think it's a good move to use the system statuses for the things they represent. For those who really don't like it, they can always customize their status icons to use the skull for Down and Out.

PR #810, opened 2025-01-11
Add option to attack roll dialog for flat bonus to the attack

### LINKED ISSUE #799: [FEATURE] Add a space for flat value input in attack HUD
**Is your feature request related to a problem? Please describe.**
In some homebrew, certain attacks allow for a flat value bonus (such as the Death's Head +1 to attacks in first party). Currently, there is no way to input a flat value bonus in the attack HUD.

**Describe the solution you'd like**
Add a space for a flat value (positive or negative) to be put in the attack HUD.

**Describe alternatives you've considered**
Remembering to manually add or subtract from the attack roll result after rolling.

**Additional context**
N/A

**Discord contact info**
skedoodlebug

eamondo2:
Adds a option to set an arbitrary attack bonus to the roll
~~Adds a config option to disable grit bonus affecting attack roll~~

Closes #799

### DIFF
--- a/src/module/apps/acc_diff/AccDiffHUD.svelte
+++ b/src/module/apps/acc_diff/AccDiffHUD.svelte
@@ -236,6 +236,14 @@
         <PlusMinusInput bind:value={base.difficulty} id="accdiff-other-diff" />
       </div>
     </div>
+    <label class="flexrow accdiff-footer accdiff-weight lancer-border-primary" for="accdiff-flat-bonus">
+      Flat Bonus
+    </label>
+    <div id="accdiff-flat-bonus" class="accdiff-grid">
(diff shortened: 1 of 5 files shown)


BoltsJ:
I've updated it so that it's _only_ the  flat modifier input and and rebased on current master, but I think the ux probably needs to be reconsidered as it makes the HUD pretty tall.

eamondo2:
Yeah the UX could use some tweaks for sure. This was mostly for a homebrew thing I'm playing in, but figured since I'd seen at least one other issue mentioning a need for a flat bonus field I'd make a PR as well.

eamondo2:
I'm also not the biggest fan of how I ended up injecting the flat bonus into the accuracy/difficulty data being passed, but it technically worked.

PR #821, opened 2025-02-23
Add flying and bolster to all core status sets

BoltsJ:
Moves towards having a unified core set of statuses


.editorconfig
Eranziel, line 1: @BoltsJ did you intend to commit this file? Not sure which editor it's for. I don't mind keeping it if you think it's useful to others.

BoltsJ: It's for [a lot of editors,](https://editorconfig.org/) but no, I didn't intend to commit it. I'll fixup the commit in a bit.

PR #822, opened 2025-02-23
Fix "Scan (Journal)" macros to make it work in foundry v12

aetherknight:
It seems like Foundry v12 finally dropped the older pre-v10 compatibility for creating the first journal page directly.

This PR:
- When creating a new scan entry, create the journal document, then add the contents as a page
- Do some cleanup of the "update an existing scan" vs "create a new scan entry" branches. _I think there's more cleanup could be done here, but I don't have a good workflow setup for producing the YAML formatting mode used by the macro from base JS text._
- Improve the handling of paragraph text in item entries. Base items benefit from the paragraph tags, but the formatting gets wonky if the GM has modified an NPC's item descriptions (eg, to add flavor text).

### DIFF
--- a/src/packs/core_macros/Scan__Journal__z0pdvwE0FitMzwRJ.yml
+++ b/src/packs/core_macros/Scan__Journal__z0pdvwE0FitMzwRJ.yml
@@ -49,10 +49,10 @@ command: >-
   const startingNumber = 1;
 
   //permissionLevel - This sets the default permission level of the scan entry
-  (from CONST.DOCUMENT_PERMISSION_LEVELS, use NONE, LIMITED, OBSERVER, or
+  (from CONST.DOCUMENT_OWNERSHIP_LEVELS, use NONE, LIMITED, OBSERVER, or
   OWNER).
 
-  const permissionLevel = CONST.DOCUMENT_PERMISSION_LEVELS.OWNER;
+  const permissionLevel = CONST.DOCUMENT_OWNERSHIP_LEVELS.OWNER;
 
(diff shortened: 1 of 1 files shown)


src/packs/core_macros/Scan__Journal__z0pdvwE0FitMzwRJ.yml
```diff
@@ -364,33 +365,38 @@ command: >-
     //This checks and updates the scan entry for the target(s) if a single scan entry exists in the specified folder for the target(s) along with the updateExisting flag.
     //If either are false then this creates a new scan entry.
 
-    let scanObj = {};
-
-    scanObj.folder = journalFolder.id;
-
     let scanEntry;
 
-    let matchingJournalEntries = journalFolder.contents.filter(e => e.name.match(actor.name));
```
aetherknight, line 373: `match` treats the `actor.name` as a regular expression. This causes naming methods that use parentheses to fail, and causes syntax errors if the actor name isn't a valid regular expression.

BoltsJ:
Please follow the [Macro Contribution Guide](https://github.com/Eranziel/foundryvtt-lancer/blob/master/docs/macro_guide.md) rather than directly editing the yaml files. Directly editing the yaml files can easily lead to invalid states in the generated compendiums.

aetherknight:
Thanks -- the guide helped me setup Foundry for local development. I ensured the macro was formatted with `foundryvtt-lancer`'s prettier config, extracted the macro, and copied back over to the src.

BoltsJ (approved):

Issue #818, opened 2025-02-12
[BUG] When importing pilot/mech sheet, Cyrillic alphabet is converted to wrong encoding


KamiliaBlow:
**Describe the bug**
When trying to import a pilot (name, callsign, bio, etc) all the Cyrillic text breaks and turns into incomprehensible characters.

**To Reproduce**
1. Create a pilot/mecha on compcon where all items that can be filled out manually will be written in Cyrillic.
2. export the json file and import it manually into foundry
3. Open a pilot or mech profile
4. Make sure that the Cyrillic alphabet is not displayed correctly

**Expected behavior**
The encoding should not break when importing a pilot file in a language with Cyrillic characters.

**Screenshots**
-

**Platform (please complete the following information):**
 - OS: Windows Server 2022
 - Browser: Foundry Light Client, Chrome
 - Foundry version v12.331
 - Lancer system version: 2.5.0 (manually compiled v12 branch)

**Additional context**
I was able to fix this for myself, maybe this will help resolve the issue faster. Changes to the src/module/actor/import.ts file

```ts
// Imports packed pilot data, from either a vault id or gist id
export async function importCC(pilot: LancerPILOT, data: PackedPilotData, clearFirst = false) {
  //new
  data.name = decodeURIComponent(escape(data.name));
  data.background = decodeURIComponent(escape(data.background));
  data.callsign = decodeURIComponent(escape(data.callsign));
  data.history = decodeURIComponent(escape(data.history));
  data.notes = decodeURIComponent(escape(data.notes));
```

and

```ts
    // Iterate over mechs
    let active_mech_uuid = "";
    for (let cloud_mech of data.mechs) {
      // Find the existing mech, or create one as necessary
      let mech = game.actors!.find(
        (m: LancerActor) => m.is_mech() && m.system.lid == cloud_mech.id
      ) as unknown as LancerMECH | null;
	  //new
	  cloud_mech.name = decodeURIComponent(escape(cloud_mech.name));
```

**Discord contact info**
kamiliablow

BoltsJ:
Temporary workaround while I investigate is to use share codes or COMP/CON login.

PR #838, opened 2025-03-11
Read C/C export as string instead of binary

BoltsJ:
Reading as binary caused problems with non-ASCII files, such as files
using Cyrillic characters.  By reading the file as text, the FileReader
is allowed to interpret the file as the appropriate encoding instead of
bruteforce reading it as ASCII.

Fixes: #818


BoltsJ:
JSON Export for checking: [cyrillic-pilot.json](https://github.com/user-attachments/files/19197400/cyrillic-pilot.json)

Eranziel (approved):
Looks like a nice fix! I'll merge later after I do my due diligence and test it myself. ;)

PR #844, opened 2025-03-17
Drop Status items on actor sheets or tokens to apply status effect

BoltsJ:
If a status type item is being created with a parent on the Item,
instead apply the status for the item and cancel the Item.  Enables
dropping status items on sheets, running the workflow so that dropping
the item applies the condition.


BoltsJ:
[Screencast From 2025-03-18 10-03-01.webm](https://github.com/user-attachments/assets/89d66487-11fc-4e56-8e59-f2186780c265)

PR #866, opened 2025-05-16
Add Simplified Chinese translation

NaricaNardica:
Add Simplified Chinese translation.
I wish you could provide more more content in en.json so that I can continue to improve the translation.
In the Chinese Lancer RPG community, there are some English terms or English nouns that are directly followed instead of being translated, such as RPV, NPC, and so on, so you may see some English letters in the Chinese translation.
Best wishes.


BoltsJ:
Distribution of translations is better done as a module I think.  This isn't really something that I can evaluate since I can't read chinese and none of the other core team members can either, to my knowledge (if I'm wrong and someone wants to be in charge of this, I'm less opposed to inclusion). Additionally, making the translation a separate module allows faster iteration than the core system and allows you to collect translations for related modules.

You can find an example that includes babele compendium translations here: https://github.com/hazmole/lancer-compendium-zhtw

NaricaNardica:
As you said, maybe it would be better to manage the Chinese translation as a mod.  
I will move forward with the Simplified Chinese translation as a module.  
In the meantime, I would be grateful if you and your team members could provide more content for translation (e.g. "TECH ATTACK", "REPAIRS" in the Mech Sheet) in the en.json of the subsequent lancer system.
Thanks a lot.

Eranziel:
Thanks for your interest! We do need to improve translation support as you say, and I hope we can iteratively improve that going forward.

PR #885, opened 2025-07-09
Apply Prettier on Foundry-v13

LukeAbby:
I noticed as I was working on updating the types that Prettier hasn't been consistently applied. The steps for this PR can be boiled down to:
1. Update Prettier - `npm add -D prettier@latest`
2. `echo "src/packs/core_macros" >>.prettierignore` and `echo "src/packs/lancer_info" >>.prettierignore`
3. `npx prettier . --write`


BoltsJ:
Prettier is part of our precommit hooks, so it's not necessary to put it in ci.  It's currently disabled in precommit for handlebars files because prettier does not support handlebars, particularly partial syntax, and because it incorrectly inserts a self-closing slash on html5 void elements as well as omitting the final newline on those files.  Furthermore, sometime in the near future I'm going to be replacing prettier with dprint (as well as adding linting) for both performance and because of the aforementioned poor html/hbs behavior.

LukeAbby:
I noticed a number of files that don't format correctly even on the current Prettier version. Though since you're planning to convert to dprint I can close this PR instead if you'd like.

PR #890, opened 2025-07-28
Foundational type fixes for v13

LukeAbby:
This PR does _NOT_ seek to bring down the number of errors to the minimum possible. A follow up PR to achieve that will be necessary. Instead this PR fixes foundational issues with the repository. The difference is that early on I got the number of errors down to the dozens but refactoring the types to be useful (i.e. avoid propogating some `any`s) has brought the errors up to 337.

There were some structural changes such as:
- Updated the `fvtt-types` dependency and made tsconfig fixes. There were a few misconfigured settings that was causing the repo to practically explode because imports like `#utils` that occur within fvtt-types. The main helpful change was `moduleResolution`; it was `node` before which is at this point ancient and doesn't support plenty of tricks. I switched to `Bundler` but `NodeNext` would've been viable in theory. 
- Turned off `skipLibCheck` to make sure errors in `.d.ts` files still get reported. I did take the liberty of converting `.d.ts` files to `.ts` files and I do plan to turn `skipLibCheck` back on but only once the repo is to 0 errors and it's been added to CI.
- Modernised the usage of fvtt-types; there's a fair bit of code that's stuck in the past in terms of configs.
- Added `type` modifiers to type-only imports.

The main focus of this PR was in data models with fixes like:
- Making sure schemas were propagated correctly. For example `ActionField` was written as simply `class ActionField extends SchemaField`. As this did not provide a schema to `SchemaField`'s generics there was no way for fvtt-types to give it a shape.
- Setting `BaseData` in each `LancerDataModel`. This models what goes on in `prepareBaseData`.
- Removing `@ts-expect-error`. I removed all `ts-expect-errors` I expect to be able to fix. That is essentially all of them.
- Removed `SystemData` as it can be derived from `Actor.SystemOfType` and friends. A follow up PR will be able to remove `SourceData` as well but I didn't want to delay this PR any further.
- Fixed some document subtype issues.
- Made `LIDField` required by default based upon the old types.


src/types/terrain-height-tools.ts
```diff
@@ -25,7 +25,7 @@ interface TerrainType {
  * LineSegments are considered equal regardless of 'direction'. I.E. p1 vs p2 order does not matter.
  */
 declare class LineSegment {
-  constructor(p1: Point | { x: number; y: number }, p2: Point | { x: number; y: number });
+  constructor(p1: Canvas.Point | { x: number; y: number }, p2: Canvas.Point | { x: number; y: number });
```
LukeAbby, line 28: Another set of missed type fixes due to being a `.d.ts` file.

src/global.ts
```diff
@@ -55,19 +58,22 @@ interface LancerInitiativeConfig<T extends string = string> {
   sort?: boolean;
 }
 
-declare global {
-  // Since we never use these before `init` tell league types that they are
-  // never undefined
-  interface LenientGlobalVariableTypes {
```
LukeAbby, line 61: `LenientGlobalVariableTypes` was moved to `AssumeHookRan`.

src/global.ts
```diff
@@ -82,30 +88,33 @@ declare global {
   // TODO: Centralize models to a single namespace
   interface DataModelConfig {
     Actor: {
-      deployable: DeployableModel;
-      mech: MechModel;
-      npc: NpcModel;
-      pilot: PilotModel;
+      deployable: typeof DeployableModel;
```
LukeAbby, line 91: The model classes are expected now.

src/global.ts
```diff
@@ -118,6 +127,13 @@ declare global {
     };
   }
 
+  namespace Hooks {
+    interface HookConfig {
+        "lancer.statusesReady": () => boolean | void;
+        "lancer.statusInitComplete": () => boolean | void;
```
LukeAbby, line 133: There's some missing custom hooks but these were some obvious ones that needed to be added.

src/lancer.ts
```diff
@@ -107,6 +107,14 @@ import handleSocketMessage from "./module/socket";
 
 const lp = LANCER.log_prefix;
 
+declare module "fvtt-types/configuration" {
+  namespace Hooks {
+    interface ApplicationConfig {
+      CombatDock: Application.Any;
+    }
```
LukeAbby, line 114: I don't know where `CombatDock` comes from. However the system expected some application hooks to exist for it. The way for fvtt-types to automatically register your hooks is this.

src/module/apps/acc_diff/data.ts
```diff
@@ -221,7 +219,7 @@ export class AccDiffHudTarget {
     };
   }
 
-  static fromParams(t: Token): AccDiffHudTarget {
+  static fromParams(t: Token.Implementation): AccDiffHudTarget {
```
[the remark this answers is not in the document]
BoltsJ: There is a system TokenDocument and Token class, so this is fine

src/module/actor/lancer-actor.ts
```diff
@@ -679,14 +644,9 @@ export class LancerActor extends Actor {
...
-  _onCreateDescendantDocuments(
-    parent: ClientDocument,
-    collection: "items" | "effects",
-    documents: LancerItem[] | LancerActiveEffect[],
-    changes: any[],
-    options: any,
-    userId: string
-  ) {
+  _onCreateDescendantDocuments(...args: Actor.OnCreateDescendantDocumentsArgs) {
+    const [parent, collection, documents, changes, options, userId] = args;
```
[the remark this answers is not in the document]
LukeAbby: You can write `_onCreateDescendantDocuments(...[parent, collection, documents, changes, options, userId]: Actor.OnCreateDescendantDocumentsArgs)` if you want. That just felt a bit messy to me was all. I can switch over if you'd like.

BoltsJ:
Is this ready to be reviewed and merged?  I can get to it today if so.

LukeAbby:
Yeah, it's ready to be reviewed @BoltsJ. There will be a follow up PR of course.

BoltsJ:
Manually merged

PR #891, opened 2025-08-03
V13/type fixes

kmoschcau:
As of 2025-08-03, this needs [this PR's branch](https://github.com/League-of-Foundry-Developers/foundry-vtt-types/pull/3453) as fvtt-types dependency to function.

I encourage reviewers to look at single commits in isolation, instead of the whole diff at the same time. I keep commits atomic for single changes, which should make the whole thing easier to review.


src/module/token.ts
```diff
@@ -114,14 +114,34 @@ export class LancerToken extends Token {
...
+    return this.document
+      .getOccupiedGridSpaceOffsets()
+      .map(o => canvas.grid?.getCenterPoint(o))
+      .filter(o => !!o);
+  }
+
+  /**
+   * This exists only to expose protected hover methods from {@link Token}.
+   */
+  onHoverIn(ev: Canvas.Event.Pointer | MouseEvent): void {
```
LukeAbby, line 126: An alternative is `LancerToken["onHoverIn"]`; this is basically an escape hatch for the TypeScript error about protected members.

Also this `Canvas.Event.Pointer | MouseEvent`; does this imply a fix is necessary to fvtt-types?

src/lancer.ts
```diff
@@ -384,8 +384,10 @@ Hooks.on("updateActor", (_actor, changes): void => {
 Hooks.on("closeSettingsConfig", () => {
   game.action_manager?.updateConfig();
 });
-Hooks.on("getSceneNavigationContext", async () => {
-  game.action_manager && (await game.action_manager!.reset());
+Hooks.on("getSceneNavigationContext", () => {
+  // FIXME: This hook has been removed for some time now. If this functionality is still needed, it needs to be placed
+  // on a different hook.
```
LukeAbby, line 389: I believe `getSceneContextOptions` will do as mentioned on Discord.

src/lancer.ts
```diff
@@ -457,9 +455,31 @@ Hooks.on("renderCombatTrackerConfig", extendCombatTrackerConfig);
 
 // Disable token vision and fog exploration by default in scene config
 Hooks.on("preCreateScene", scene => {
-  scene.updateSource({ tokenVision: false, "fog.exploration": false });
+  scene.updateSource({ tokenVision: false, fog: { exploration: false } });
```
LukeAbby, line 458: Yeah I'd like to add back dotkey support but last time I did it I lagged out repos so bad. Probably some dumb data model stuff.

LukeAbby (commented):
Obviously a draft PR but I did just want to give a quick review anyways.

BoltsJ (commented):
Quick review with some requests before you get too far in.

BoltsJ:
Also, I find the "conventional commits" style to make it hard for me to read the commit log, and it's not the style used for the repo.

src/module/token.ts
kmoschcau: The MouseEvent is just there to make calling the methods easier. Since they were called from native DOM-Events. The canvas pointer event requires a lot more properties to be added, which are not even used in the two token-methods. `_onHoverOut` doesn't even use the event at all. I'd actually argue we can widen the parameter types in the types for these two methods. Yes, they are documented as being full custom canvas pointer events, but they are not used as such.

src/module/helpers/dragdrop.ts
```diff
@@ -260,33 +273,33 @@ export async function resolveNativeDrop(drop: string | FoundryDropData): Promise
         document,
       };
     }
-  } else {
+  } else if ("type" in drop) {
```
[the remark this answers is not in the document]
kmoschcau: They might not be necessary at runtime, because you already know that this property will definitely be there. But if our goal is to make the typescript errors go away, these checks are the easier alternative to having to ensure narrower types in more complex ways. This is fairly standard typescript 101 for runtime type checks.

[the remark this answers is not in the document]
kmoschcau: From my understanding this is a system specific enhancement. It's not part of core.

kmoschcau:
@LukeAbby I would like your opinion on https://github.com/Eranziel/foundryvtt-lancer/pull/891/commits/a09ad703a59383de006509d3b1fc3db5f5403878.

PR #919, opened 2026-03-30
Migrate to svelte 5

BoltsJ:


BoltsJ:
Styling issues was caused by vite.  It's worked around by using vite-plugin-foundryvtt.  The only issue left is that the first time you try to open a hud application, it silently ignore the button click.  Not sure what's causing it, probably the appv1 sheets

PR #927, opened 2026-04-12
Fix meltdown text and engineering check

### LINKED ISSUE #918: [BUG] Overheating Automation Errors
**Describe the bug**
There are two problems with the Meltdown result in the overheating automation.
1) The description says you can retry the Engineering check as a free action. 
https://github.com/Eranziel/foundryvtt-lancer/blob/9fddbe556c39668a668999315c80fbd2a91ffd5c/public/lang/en.json#L416
Comp/Con says it's a full action to retry the check. An older PDF apparently had this mistake and was later corrected.
Image

2) The prompt in the chat card doesn't include the actual engineering stat. 
Image

**To Reproduce**
1. grab a mech and reduce it to 3 out of 4 stress
2. overheat
3. roll a one
4. click the prompt in chat

 - Foundry v12.343
 - Lancer system v2.11.1

tw-blob:
Fixes #918

### DIFF
--- a/src/module/flows/overheat.ts
+++ b/src/module/flows/overheat.ts
@@ -284,7 +284,7 @@ export async function overheatInsertEngCheckButton(
     state.data.embedButtons.push(`<a
       class="flow-button lancer-button"
       data-flow-type="check"
-      data-check-type="engineering"
+      data-check-type="eng"
       data-actor-id="${actor.uuid}"
     >
       <i class="fas fa-dice-d20 i--3"></i> ENGINEERING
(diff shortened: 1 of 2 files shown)


src/module/flows/overheat.ts
```diff
@@ -284,7 +284,7 @@ export async function overheatInsertEngCheckButton(
     state.data.embedButtons.push(`<a
       class="flow-button lancer-button"
       data-flow-type="check"
-      data-check-type="engineering"
+      data-check-type="eng"
```
Eranziel, line 287: This isn't quite the right fix. The change is needed in `lancer.ts` on line 591, which is the listener that gets attached to this button.

Eranziel (approved):
Thanks for working on this! As noted below it's not quite the right fix, but it put me on track. I'll take it from here, thanks!

PR #932, opened 2026-05-04
Fixed typo in Scan__Journal__z0pdvwE0FitMzwRJ.yml

AshaDasha:
Fixed typo in comment on line63 from:

of a token (from CONST.DOCUMENT_PERMISSION_LEVELS, use NONE, LIMITED,

to

of a token (from CONST.DOCUMENT_OWNERSHIP_LEVELS, use NONE, LIMITED,


BoltsJ:
This macro has been reworked to use a system flow, so this change is presumably no longer needed.

PR #934, opened 2026-05-11
Add CCv3 pilot import compatability

annoyedb:
(assuming CORS errors are fixed on CC's end)
---
> `packed-types.ts` updated with new CCv3 data model requirements and maintained to be compatible with CCv2

Some packed data types had bigger changes than others, but most of them were just additions to their data model. Other changes were mostly Types changes/fixes for better consistency. It should still be compatible with CCv2 data.

> `enums.ts` updated to be closer aligned to enums used by COMP/CON. Some unused enums were repurposed for more consistency

The changes are mostly capitalization changes, which would only break hard string comparisons; but I double checked the ones that I changed for that specific scenario and I'm happy to say that there were none. Other changes include hijacking the `Duration`, `PilotEquipType` (renamed to `PilotItemType`), `EncounterSide` (renamed to `Target`) enums since they weren't being used anywhere. 

> `pilot.ts`, `pilot_armor.ts`, `pilot_gear.ts`, `pilot_weapon.ts`, `item.ts`, `bonus.ts` updated to account for changes from required keys turning into optional keys

> `pilot-sheet.ts `added new filter for reading CCv3 share codes

> `converter.ts` added a few sample cases for a few new tags added by CCv3. And also some minor formatting changes

This is absolutely *not* all of the new tags added in CCv3 (`tg_set_max_uses`, `tg_set_damage_type` for the Engineer talent alone) but it felt out of scope for this PR. This was meant to be more of a footnote for when *LCP* importing is updated to support the current CCv3 standard. There is likely to be a lot of warnings that will pop up for many people importing v3 pilots due to this, but I really, *really* did not feel like parsing through all of the LCPs for this PR.

> `compcon.ts` added CCv3 share-code fetching

> `import.ts` added CCv3 data handling on importing pilots and added importing reserves to CCv2 data handling

I subscribed to the 'if it ain't broke don't fix it' philosophy for the actual importer code. `updatePilot`, `updateMech`, `clearPilotEmbeddedDocuments`, `hasCreatePermissions` and `getActorItemByLid` can replace many spots in the v2 importer, but I've only thoroughly tested them for the v3 importer.

I decided to attempt to create items from the compendium and then updating it with the imported data *before* trying to create documents on their own. It's a little redundant, but it seemed like a decent guard in case there is something that may be missed that could only be extracted from the LCP. As designed by the new CCv3 spec, however, pilot data now includes everything required to build the pilot in the moment, and from my testing with a handful of both official and unofficial LCPs, there were no issues (other than unknown tags added from bonuses)

> `requests.ts` changed mech deployables to append the mech's name rather than the pilot's callsign to better distinguish instances where multiple mechs (and mech loadouts) under the same Lancer may have the same deployable systems/weapons

Admittedly, this is unrelated to importing pilots, but it was really bothering me in play and in testing :sob:

> Other TS warning fixes probably


src/module/actor/pilot-sheet.ts
```diff
@@ -70,15 +71,26 @@ export class LancerPilotSheet extends LancerActorSheet<EntryType.PILOT> {
...
+            ui.notifications!.info("Importing character from V3 share code...");
+            console.log(`Attempting import with V3 share code: ${pilot.system.cloud_id}`);
             try {
-              raw_pilot_data = await fetchPilotViaShareCode(pilot.system.cloud_id);
+              raw_pilot_data = await fetchV3PilotViaShareCode(pilot.system.cloud_id);
             } catch (error) {
-              ui.notifications!.error("Error importing from share code. Share code may need to be refreshed.");
-              console.error(`Failed import with share code ${pilot.system.cloud_id}, error:`, error);
+              ui.notifications!.error("Error importing from V3 share code.");
+              console.error(`Failed import with V3 share code ${pilot.system.cloud_id}, error:`, error);
```
Eranziel: Nitpick: looks like I never added the Lancer log prefix in this file. Could do that in this PR.

src/module/enums.ts
```diff
@@ -288,12 +286,32 @@ export enum SystemType {
...
-  Turn = "Turn",
-  NextTurn = "NextTurn",
-  Scene = "Scene",
-  Mission = "Mission",
+  NextTurnStartSelf = "next_turn_start_self",
+  NextTurnEndSelf = "next_turn_end_self",
+  NextTurnStartTarget = "next_turn_start_target",
+  NextTurnEndTarget = "next_turn_end_target",
+  // RoundStart = "round_start_",
+  // RoundEnd = "round_end_"
```
Eranziel: Are round start and round end not used in LCPs yet?

src/module/enums.ts
```diff
@@ -288,12 +286,32 @@ export enum SystemType {
...
+}
+
+export enum Target {
```
Eranziel: Suggestion: rename to TargetDisposition, to make it more specific and align with Foundry's combatant dispositions.

src/module/actor/import.ts
```diff
@@ -14,25 +14,894 @@ import {
...
+      name: data.name,
+      texture: {
+        src: replaceDefaultResource(pilot.prototypeToken?.texture?.src, portrait, pilot.img),
+      },
+    },
+  });
+}
+
+/**
+ * Updates the Pilot document, given COMP/CON data.
```
Eranziel: Mech here, not pilot. ;-)

src/module/actor/import.ts
```diff
@@ -14,25 +14,894 @@ import {
...
+  });
+
+  // @ts-ignore Please stop yelling at me :sob:
+  return foundry.applications.api.DialogV2.prompt({
+    window: {
+      title: `Select Pilot Loadout`,
+      icon: "fas fa-triangle-exclamation",
+    },
+    content: `
+      <span>Multiple pilot loadouts found. Please select a single loadout to import and use:</span>
```
Eranziel, line 279: Ooh, nice! Thanks for implementing this!

src/module/actor/import.ts
```diff
@@ -14,25 +14,894 @@ import {
...
+  const data = importedData.data;
+  if (!data) {
+    console.error(`${lp} Tried using CCv3 importer on CCv2 data`, importedData);
+  }
+
+  if (clearFirst) {
+    await clearPilotEmbeddedDocuments(pilot);
+  }
+
+  hasCreatePermissions();
```
Eranziel, line 348: Hmm, I thought we aborted the import if the user didn't have actor create permission. Did that get lost in the refactor, or am I dreaming?

Eranziel (changes requested):
Thanks so much for working through this! Looks pretty good on a static read through, these are mostly nitpicks. I'll try to test this branch out soon to see if I can catch anything else.

src/module/enums.ts
annoyedb: They are, but they are terminated by some number

E.g. `round_start_3` or `round_end_5`

So you would need to read the actual string

This AFAIK is completely new in ccv3 / it isn't being utilized by the system at current. I suppose I can add a helper function to parse the tag but it would be unused anywhere

src/module/actor/import.ts
annoyedb: I'm not happy about the ts-ignore but I wasn't sure if the types package was out dated or something 😬 

I'm not really sure how to resolve it

src/module/actor/import.ts
annoyedb: https://github.com/Eranziel/foundryvtt-lancer/blob/761e50320a84970113267068b584fae0f5c403ca/src/module/actor/import.ts#L66

At least in this it doesnt look like it

Admittedly I didnt test this because of the fact I just copy and pasted it verbatim 😬

annoyedb: I can add a return there and to the v2 importer if thats preferable

annoyedb: On a second read, I believe the original intent was that users without creation permissions could still try to *update* their pilot, so import is not aborted

src/module/enums.ts
annoyedb: Added duration tag validator/extractor and uncommented the enum keys for possible future use
https://github.com/Eranziel/foundryvtt-lancer/pull/934/changes/53951224068281f4e2a51e4b821de602df2509cc

src/module/actor/import.ts
annoyedb: In any case I suppose I can take a look back at it once the overall types issues are resolved in the repo

src/module/actor/import.ts
Eranziel: Ah yeah, that makes sense.

src/module/actor/import.ts
```diff
@@ -14,25 +14,898 @@ import {
...
+          (
+            await pilot.createEmbeddedDocuments("Item", [
+              {
+                type: EntryType.SKILL,
+                name: item.data.name,
+              },
+            ])
+          )[0].id;
+
+        const ccData = unpackSkill(item.data, _context);
```
Eranziel: This consistently sets the skill rank to 1 because `unpackSkill` is hard coded to do so. It was written under the assumption that it was always consuming LCP data to be used to make items for compendiums.

src/module/actor/import.ts
```diff
@@ -14,25 +14,898 @@ import {
...
+/**
+ * Gets an actor's Item by their LID and tries to insert the item into their Document.
+ * Searches from the item pool first before attempting to look up the Compendium entry.
+ * @param lid           - LID to lookup
+ * @param actor         - Actor to insert the embedded documents into
+ * @param itemPool      - Item pool to reference
+ * @param missingItems  - Array appending an object holding the `actor` and `lid` of any failed lookups
+ * @return The `LancerItem` if successful or `null`
+ */
+async function getActorItemByLid(
```
Eranziel: Suggested rename for clarity: `getOrCreateActorItemByLid`. I was doing some debugging and got confused, I thought this would only get existing items.

src/module/actor/import.ts
```diff
@@ -14,25 +14,898 @@ import {
...
+            ])
+          )[0].id;
+
+        gears.push(id);
+
+        const ccData = unpackPilotGear(item.data as PackedPilotGearData, _context);
+        pilotItemUpdates.push({
+          ...ccData,
+          _id: id,
+        });
```
Eranziel: I think it's a bit redundant here to add an item to an actor from the compendium and then replace the data with what was in the import JSON anyway. If the item was in a compendium and we pulled it from there, I think that should be sufficient. 🤔 Thoughts?

If we change approaches here, the same change should be made to the handling of all item types.

src/module/actor/import.ts
```diff
@@ -14,25 +14,898 @@ import {
...
+        _id: id,
+        system: {
+          ...ccData.system,
+          curr_rank: item.rank,
+        },
+      });
+    }
+
+    // Bonds
+    if (data.bond) {
```
Eranziel: If the pilot doesn't have a bond, this section logs a warning and shows a persistent warning notification to the user. This can all be skipped IMO if there is no bond item LID.
```suggestion
    if (data.bond?.bondId) {
```

src/module/actor/import.ts
```diff
@@ -14,25 +14,898 @@ import {
...
+        (
+          await pilot.createEmbeddedDocuments("Item", [
+            {
+              type: EntryType.LICENSE,
+              name: item.stub.name,
+            },
+          ])
+        )[0].id;
+
+      const ccData = unpackLicense(item.stub.name, lid, item.stub.source, _context);
```
Eranziel: Same as skills, licenses always import as rank 1 because `unpackLicense` assumes it's always consuming LCP data.

src/module/actor/import.ts
```diff
@@ -14,25 +14,898 @@ import {
...
+          type: mount.mount_type as MountType,
+          slots: populatedSlots,
+        });
+      }
+
+      // Update all items
+      await mech.updateEmbeddedDocuments("Item", mechItemUpdates);
+
+      // Perform base mech update
+      await updateMech(mech, pilot, importedMech, ownershipLevel, populatedMounts, populatedSystems, compendiumFrame);
```
Eranziel, line 856: Freshly imported mechs have 0 hp, structure, stress, and repairs. I think this would be the right place to update the value of those stats to max.

Imported deployables also don't get their hp set to max... not sure where would be best to handle those. Maybe here as well? It's late here, I'm getting fuzzy.

src/module/actor/import.ts
```diff
@@ -14,25 +14,898 @@ import {
...
+                },
+              ])
+            )[0].id;
+
+          const ccData = unpackMechWeapon(weaponSlot.data, _context);
+          mechItemUpdates.push({
+            ...ccData,
+            _id: weaponId,
+            system: {
+              ...ccData.system,
```
Eranziel: Not sure which part is causing it, but weapon "on hit" effects aren't being parsed right.

Image

src/module/actor/import.ts
```diff
@@ -14,25 +14,898 @@ import {
...
+          if (!extraSlot.weapon) continue;
+          const { weapon, weaponId, mod, modId } = await processMechWeapon(extraSlot.weapon);
+          populatedSlots.push({ mod: mod ? modId : null, weapon: weapon ? weaponId : null, size: extraSlot.size });
+        }
+
+        populatedMounts.push({
+          bracing: mount.lock ?? false,
+          type: mount.mount_type as MountType,
+          slots: populatedSlots,
+        });
```
Eranziel, line 849: I might not be commenting on the right place, but mount types modified by a core bonus (Mount Retrofitting) don't show with the correct type after import. This is pretty minor IMO.

Eranziel (changes requested):
I had a chance to play with the importer a bit tonight and noticed a couple bugs. To be expected, there's so many edge cases to handle.

src/module/actor/import.ts
annoyedb: Ah, I didn't notice that the skill level was incorrect. `item.data` should have the `rank` key, I'll have it fixed

src/module/actor/import.ts
annoyedb: I could have sworn I caught this one specifically, but I'll have it fixed as well

src/module/actor/import.ts
annoyedb: I'll have to check the discrepancies with V2 and V3 data, I have a feeling the tag probably changed

src/module/actor/import.ts
annoyedb: That's... weird. My test pilot imports all mechs with their HP correctly set. Could I have the JSON of what you're testing with? I use these two: 

[LANCER_v3.json](https://github.com/user-attachments/files/27745019/LANCER_v3.json)
[LANCER2_v3.json](https://github.com/user-attachments/files/27745018/LANCER2_v3.json)

src/module/actor/import.ts
annoyedb: As I've mentioned, it really is redundant

But I chose to do it this way because I saw it as a natural guard prioritizing the system's interpretation and sanitization of LCPs. That said, I'm not entirely confident in my knowledge of how LCPs get imported by the system (edge case scenarios: formatting, typos, non-existent types, etc) versus how CC would import LCPs and then take the additional step of exporting them (does it automatically sanitize fields? what are its own edge cases like?).

There's also some things such as tags (limited, arcing, smart, etc) that aren't packaged with the exported pilot and potentially more that *could* be missing from the export as well.

You could argue that removing the redundancy would be a performance benefit, but... 🤷 

Actually, I suppose there's also version discrepancies between LCPs to potentially consider. 

Honestly, not sure how to handle it lol

Eranziel: My issue here is (unless I've misread what the code does) it imports the item from the compendium and then overwrites the data with what the pilot JSON has. That makes the compendium data kind of pointless.

I'd prefer to use only the compendium data if it exists (basically same as v2 imports). If the item can't be found in the compendium, then unpack the data from pilot JSON and use that. 

There may be some data from the pilot JSON that we want to import either way, but it should be fairly limited. I'm thinking play state stuff like limited uses remaining, core power spent, etc...

src/module/actor/import.ts
Eranziel: Here's the two I was using:

[AMAROK.json](https://github.com/user-attachments/files/27771960/AMAROK.json)
[NAMAZU.json](https://github.com/user-attachments/files/27771970/NAMAZU.json)

src/module/actor/import.ts
Eranziel: Oh, these probably changed from a simple string to a set of effects, didn't they?

src/module/actor/import.ts
annoyedb: Okay, not a problem

I'll refactor it to remove the overwrites. So you would prefer to prioritize LCP/compendium data over the import data?

src/module/actor/import.ts
annoyedb: I don't recall, I'll have to double check this later today or tomorrow

src/module/actor/import.ts
annoyedb: I can confirm that it is an issue with CC's v3 exporter. I don't know when those were made, but you can see in both files that the `stats` object for the mech (not the frame) is zeroed out. It seems to be resolved when I re-export that mech after importing it into CC, however I am having another issue right now where that object is also zeroed out for the last exported mech if the pilot has multiple mechs.

Either way, I think it is safe to say that is a CC issue for now

src/module/actor/import.ts
annoyedb: While it is an unsatisfactory answer, I can confirm that the CCv2 importer does the same thing.

Though while looking into it, I will note that the way it is currently being handled is incorrect. Both CCv2 and CCv3 data shows which mount is being retrofitted via the `bonus_effects` array under each mount. [These lines should be refactored to reflect that](https://github.com/Eranziel/foundryvtt-lancer/blob/6725324cbc905372475b164aa71cc33a1771612e/src/module/actor/loadout-util.ts#L362) or [perhaps adding a case handler to `converter.ts`](https://github.com/Eranziel/foundryvtt-lancer/blob/6725324cbc905372475b164aa71cc33a1771612e/src/module/effects/converter.ts#L475)

Eranziel: That's fine, and tbh unsurprising. Mount handling has always been a bit junky, so I've just left it alone because it works well enough. 

A refactor is a good idea, and IMO should be left out of this PR for scope.

src/module/actor/import.ts
Eranziel: That's my current thoughts, but we can discuss more on Discord if you like.

src/module/actor/import.ts
annoyedb: I... don't know what the issue is 
image

I'll just assume I accidentally fixed it or something

Eranziel (approved):

PR #946, opened 2026-05-26
Add Plain Font Setting For Accessibility

Shteb:
This feature adds a "Simple/Plain Font" accessibility toggle in the system settings which allows a user to replace all sci-fi fonts with Helvetica. The pause menu is excluded because it looks rather awkward without the stylish text.

Made following some mentions of individuals with dyslexia struggling to read the new fonts.
image
image

### DIFF
--- a/src/lancer.scss
+++ b/src/lancer.scss
@@ -105,4 +105,14 @@
   /* Elements Layer */
 
   /* ---------- */
+
+  /* Simple fonts override */
+  body.lancer-simple-fonts {
+    --header-font-family: Helvetica, Arial, sans-serif;
+
+    /* Pause overlay is intentionally excluded from the simple fonts setting */
+    #pause.lancer-pause figcaption {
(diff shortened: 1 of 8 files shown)


src/module/integrations/lancer-combat-tracker-dock.scss
```diff
@@ -73,7 +73,7 @@
 
         p,
         ul {
-          font-family: Consolas;
+          font-family: var(--monospace-font-family);
```
Eranziel, line 76: Good catch!

src/lancer.scss
```diff
@@ -105,4 +105,14 @@
   /* Elements Layer */
 
   /* ---------- */
+
+  /* Simple fonts override */
+  body.lancer-simple-fonts {
```
Eranziel, line 110: FYI I'm going to move these rules to the overrides layer. If it's for accessibility (valid), it should take priority over any other styling.

Eranziel (approved):

PR #948, opened 2026-06-02
Foundry v14 support

BoltsJ:
- [x] Fix deprectation of DataField#migrateSource breaking custom fields
- [x] Remove custom region containment code
- [ ] ActiveEffects 2.0 support
  - [ ] Add custom phases for system effects
  - [ ] Pass change objects directly instead of using temporary documents
  - [ ] Remove EffectHelper in favor of a more vanilla and types-friendly way of passing effect data
  - [ ] Consider whether our changes to active effects require a customized ActiveEffect class
  - [ ] Use the token active effect system to manage size automation
  - [ ] Deprecate Status/Condition items in favor of an ActiveEffect subtype
  - [ ] Consider adding status duration automation
- [ ] Consider moving LoadoutHelper functionality into LancerActor or refactoring it to a mixin on relevant TypeDataModels to reduce circular types
- [ ] Consider the same for StrussHelper
- [ ] Use database operation batching instead of update chaining for NPC class application and pilot imports
- [ ] Replace MeasuredTemplates with scene regions
  - [ ] Add system defaults to scene controls for template regions
  - [ ] Use an emanation shape for burst templates
- [ ] Replace TinyMCE editors with ProseMirror editors


binarydesu:
I have been running v14 against a live world. I opened #964 with five regressions I hit. All of them are silent: no exception, just wrong or missing behavior. Two of them break this branch today. `prepareBaseData` never chains to `super`, and the code passes `CHAT_MESSAGE_STYLES` as ChatMessage `type`.

Before I go further on the ActiveEffects 2.0 items, I would rather ask than guess at your design.

**1. "Pass change objects directly instead of using temporary documents."** Do you mean an override of `applyActiveEffects()`, so inherited changes go straight through the static `ActiveEffect.applyChange(doc, change, {replacementData})`? Today `EffectHelper.inheritedEffects()` builds `LancerActiveEffect` documents only so the core loop has something to iterate. I first assumed those temporary documents were unavoidable, because `allApplicableEffects()` yields documents. The static entry point looks like it makes them unnecessary.

**2. "Add custom phases for system effects."** Are the priority constants in `converter.ts` meant to become registered phases? I mean `FRAME_STAT_PRIORITY` through `FEATURE_OVERRIDE_PRIORITY`. The v14 `CHANGE_PHASES` API looks aimed at the ordering problem those numbers solve by hand today.

**3. Is `EffectHelper` public API?** `effectHelper` is public on `LancerActor` and reachable through `game.lancer`. Also, `setEphemeralEffects`, `clearEphemeralEffects` and `findEffect` have no callers inside the system. If modules depend on them, then "remove EffectHelper" is a breaking change. That is worth settling before anyone starts.

One more thing, separate from the questions above. `clearEphemeralEffects` used the `"-="` deletion key, which v14 silently ignores. The update resolves without an error and the effects stay put. Assigning `null` works, because the field is already nullable with an initial of `null`.

I am happy to take any of these on. I just do not want to guess wrong about the shape you have in mind.

PR #953, opened 2026-06-15
Added infect as a damage type mirroring burn, but for heat. Fixed sta…

yiatz:
…tFlow switch to use correct labels.


yiatz:
Unintentional PR from fork.

Issue #952, opened 2026-06-14
[BUG] The "Cost" property is ignored when activating limited systems and weapons.


yiatz:
When a weapon or system is activated the flow is hard coded to use a cost of "1" rather than using the "cost" property on the weapon or system. This causes certain weapons with optional limited costs to always consume a charge and not be able to attack when no charges are left. Example from official content is the "Superthermal Blade" from "Operation Winter Scar". This weapon has an optional on hit effect of "... you may expend a charge to to deal +3 heat...". While this weapon is not configured with a "cost" of 0 (it's undefined), I have tested with 0 as the value and verified in the code that this value is ignored.

You can see why this value is ignored in item-utils.ts line 170:
```typescript
if (state.item.isLimited()) itemChanges.uses = { value: Math.max(state.item.system.uses.value - 1, 0) };
``` 
All the limited use flows seem to use a magic number for the cost rather than pulling the value from the weapon or system. It looks like the cost field was intended to be used from the comment in packed-types.ts line 816:
```typescript
cost?: number; // How many limited uses to consume per firing?
```

The LCP editor has a field for this cost on weapons that defaults to null, but does properly fill the json field and value when a non-null value is set. See image below. I don't see a corresponding field when editing a system.

**To Reproduce**
non-zero charges left:
1. Deploy a mech with the Superthermal Blade on a mount.
2. Verify there is at least one charge remaining on the weapon.
3. Roll Attack with the Superthermal Blade.
4. Click "Roll" to perform the attack.
5. One use will be consumed.

zero charges left:
1. Deploy a mech with the Superthermal Blade on a mount.
2. Remove all charges for the weapon.
3. Click to Roll an attack with the weapon.
4. An error message will appear stating "Mech Weapon Superthermal Blade has no remaining uses!"

**Expected behavior**
I believe the intended use for the "cost" property would be to set this to 0 on weapons like the Superthermal Blade so that they can be activated without consuming charges and allow activation when there are zero charges remaining.

**Screenshots**
<img width="512" height="136" alt="Image" src="https://github.com/user-attachments/assets/a188419a-ee52-4b06-8a5b-d9185f865af8" />

**Platform (please complete the following information):**
 - OS: Windows
 - Browser firefox
 - Foundry version 13 stable
 - Lancer system version 3.0.0 build 351

**Additional context**

**Discord contact info**
[...]

PR #955, opened 2026-06-16
Issue 952

yiatz:
Evaluate charge cost based on the action. Relates to https://github.com/Eranziel/foundryvtt-lancer/issues/952

### DIFF
--- a/src/module/flows/item-utils.ts
+++ b/src/module/flows/item-utils.ts
@@ -84,7 +84,10 @@ export async function checkItemLimited(
     // The frame is not limited, so we're good.
     return true;
   }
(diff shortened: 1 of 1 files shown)


yiatz:
The current logic assumes that any use costs 1 charge for limited systems and weapons. This change will use either the action cost (system, weapon, action) or default to 1 if a value can't be found. The only difference in functionality is that when weapon actions are imported it gives them a default cost of 0 if one isn't defined. It's possible to remove the default value so that it would continue to work as it does now, treating undefined as 1 for weapon actions.

Eranziel (approved):

Issue #954, opened 2026-06-15
[BUG] Weapon actions on a weapon with a single profile duplicate the action


yiatz:
**Describe the bug**
If a weapon with a single profile has an action associated with it, it will consider that both a profile action and a system action. This causes the ability to be duplicated in the UI. This is because the json data has the action under the system (weapon).

This issue is not present when there is a second profile. When there are multiple profiles the json data has the actions under the specific profiles. There's currently not a way to add an action outside of profiles in the current LCP editor.

<img width="167" height="189" alt="Image" src="https://github.com/user-attachments/assets/5be78015-99b0-4de7-8868-50ea0479f57f" />

~~I don't think this feature is used in first party content. If I can find a first party example I'll add it later.~~
first party example of this is the Autogun:

<img width="665" height="311" alt="Image" src="https://github.com/user-attachments/assets/056f1d81-57a6-4515-adbd-091482b35cfb" />

**To Reproduce**
1. Equip a weapon with a single profile and an attached ability.
2. Open the Mech Loadout page.
3. The weapon action(s) will be duplicated.

**Expected behavior**
The actions should be imported in to a single location rather than duplicated. This will fix the UI portion of the bug as well.

**Screenshots**

**Platform (please complete the following information):**
 - OS: Windows
 - Browser firefox
 - Foundry version 13 stable build 351
 - Lancer system version 3.1.1

**Additional context**
I tried manually editing an LCP to have a weapon with an array of one profile containing the ability. Import of that LCP failed.

**Discord contact info**
If you are willing to answer questions regarding this bug report on Discord, or if you want to be notified when it is fixed, please leave your Discord ID here.
yiatz or Yiatz#9443

PR #956, opened 2026-06-16
Remove printing base actions on weapons.

yiatz:
Relates to https://github.com/Eranziel/foundryvtt-lancer/issues/954

### DIFF
--- a/src/module/helpers/item.ts
+++ b/src/module/helpers/item.ts
@@ -718,7 +718,7 @@ data-action="set" data-action-value="(int)${i}" data-path="${weapon_path}.system
   let on_crit = profile.on_crit ? effectBox("On Crit", profile.on_crit) : "";
 
   // Generate actions
-  let actions = weapon.system.actions.length > 0 ? buildActionArrayHTML(weapon, `system.actions`) : "";
+  let actions = "";
   actions +=
     profile.actions.length > 0
       ? buildActionArrayHTML(weapon, `system.profiles.${weapon.system.selected_profile_index}.actions`)


yiatz:
The current implementation of weapon actions duplicates the action in both the weapon and the profile when there's only one profile. When there are multiple profiles then all actions are in the profiles. Pulling the actions from the weapon will only duplicate actions.

This could change in the future if the LCP editor is changed to allow actions to be added to the weapon outside of a profile.

src/module/helpers/item.ts
```diff
@@ -718,7 +718,7 @@ data-action="set" data-action-value="(int)${i}" data-path="${weapon_path}.system
   let on_crit = profile.on_crit ? effectBox("On Crit", profile.on_crit) : "";
 
   // Generate actions
-  let actions = weapon.system.actions.length > 0 ? buildActionArrayHTML(weapon, `system.actions`) : "";
```
Eranziel, line 721: This change alone isn't sufficient because it causes another issue - actions on the weapon itself (not belonging to a profile) get hidden. For example, see the D/D 288.

I think instead we need to not add profile actions to the `system.actions` during data prep. Probably still useful to have a collection of all actions from the weapon + profiles, but move that to a new derived data field like `all_actions`.

Eranziel (changes requested):

src/module/helpers/item.ts
yiatz: Let me take another look at this. I was hesitant to play with the existing fields because I didn't want a breaking change. I'll make it so that it populates two areas: a field containing all abilities "all_actions" and abilities per profile. Then we can have this section of code just load from the profile.

Eranziel: The root cause here is that `weapon.system.actions` already includes both the actions on the weapon itself (like D/D 288) and profile actions, but then the next statement adds profile actions again.

I'm sure the profile actions are being added to `system.actions` during data preparation. What I can't recall is whether actions from *all* profiles are being added at that step (so it's a collection of all possible actions for the weapon), or if only the active profiles actions are added. Depending on which is the case, I think the best course would be:
1. If only active profile actions are added to `system.actions`, then remove the statement after this one. The work is already done, all we need to do is render the actions in `system.actions`.
2. If all profile actions are added to `system.actions`, it becomes more in depth. I think in that case we should create a new derived field on weapons (`system.all_actions`) and put that data there. `system.actions` should be refactored to contain either only base actions, or base + active profile. Probably the latter since that's closer to the  current behavior. Then make the same change in this file as I mentioned for 1.

2 is a breaking change indeed, but I'd say it would actually be fixing an unintended oversight and overall be an improvement. How does that sound?

PR #964, opened 2026-07-22
Fix five silent v14 regressions

binarydesu:
Builds on #948 (`foundry-v14`).

Five bugs found by running v14 against a live world. Three produced no error at all. The other two showed only a console error while the flow reported success, so nothing surfaced in the UI:

- **Lancer actors failed data preparation.** `prepareBaseData()` chained to `super` only on its early-return path for non-Lancer types. So v14 never initialized the new `tokenActiveEffectChanges`, and `applyActiveEffects()` threw on it.
- **Chat cards never posted.** The code passed `CHAT_MESSAGE_STYLES` as ChatMessage `type`. That field is the document subtype now, so v14 rejects the create. This is `createChatMessageStep`, so it covers every flow card: attack, damage, tech, structure, overheat and the rest. Moved to `style`.
- **The popout description editor erased text.** It renders through the TinyMCE-era `{{editor}}` helper, which no longer fills in a value. The dialog opened blank and saved that blank back. Rerouted through the existing `richTextEdit()` (ProseMirror).
- **Blind, Private and Self labels stopped appearing.** Foundry v14 removed the `CHAT.RollBlind`, `CHAT.RollPrivate` and `CHAT.RollSelf` keys, so `localize()` put the key name itself on the card. The label now comes from `CONFIG.ChatMessage.modes[key].label`, which is the same string the chat controls show. A correction to my first attempt at this: the setting still returns the legacy values in v14. The "Blind to Gamemasters" control sets `core.rollMode` to `"blindroll"`, not to `"blind"`. The removed keys were the whole cause. The switch also accepts the current spellings, in case a later version changes the stored value.
- **Weapon damage and range bonuses never applied.** These went out as a change carrying an out-of-range numeric mode (12), which v14 discards, so `system.bonuses.weapon_bonuses` stayed empty (Neurolink Targeting, Gyges Frame). They now go out as `custom` changes, which still reach the `applyActiveEffect` hook. v14 hands the handler an already-resolved `value`, so the handler parses it only when it is still a string.

The remaining commits move deprecated globals onto `foundry.appv1.api.*` and `foundry.applications.ux.*`, replace `bringToTop` with `bringToFront` for the ApplicationV2 sheets, and drop the removed `CONFIG.ActiveEffect.legacyTransferral` and status-effect `icon` fallback. This does not catch every deprecated global: bare `renderTemplate` calls remain in `flows/stabilize.ts` and `flows/text.ts`.

**Testing.** I verified the first seven commits by hand in Foundry 14.365: LCP import, COMP/CON pilot import (share code and JSON), attack, damage and structure flows, status application, and the weapon-bonus fix. For the eighth commit I set each of the three modes from the chat controls in a live v14 world and rolled. The cards read "Blind to Gamemasters", "Private to Gamemasters" and "Self Only", with the expected `blind` flag and whisper list. The same rolls before the change put the bare key names there. A console check in that world also confirmed two more things. Every actor prepares data without error. A flow card carries `style` 2 with `type` `"base"`. `tsc --noEmit` stays at the 503 errors this branch already has. None are new.

**Partial.** The TinyMCE item on your checklist is not closed. The live popout path now uses ProseMirror, but `templates/window/html_editor.hbs` still holds the only `{{editor}}` call and is still preloaded, and `HTMLEditDialog` is left unreferenced.

**Not included:** the `CONST.ACTIVE_EFFECT_MODES` to string-type migration. It touches about 90 sites and costs 20 new `tsc` errors, because fvtt-types has no v14 release and rejects the `type` field. Back-compat runs to v16, and ActiveEffects 2.0 will rewrite that code anyway.

---

*Disclosure: written with Claude (Anthropic). Claude did the investigation and the edits. I drove it and tested every change in a live v14 world. For the eighth commit Claude ran the checks against that world itself, through the browser console. Nothing here went in without a runtime check first. I found all five bugs above by playing the system, not by reading release notes.*


Eranziel:
Hi there. I appreciate the enthusiasm and desire to pitch in, and also appreciate the honesty about the source of these changes. However, I do not plan to accept AI-written contributions to this repo. I have very limited time to work on this project, and I don't want to spend that time reviewing AI code.

At a cursory glance, even some of the simple changes here take an objectively wrong approach. I do not wish to spend my energy guiding an AI instead of getting to actually create something myself. Not to mention the future tech debt by accepting poor (or poorly understood) code to the repo.

I will gladly accept human-written code which is well understood by the author so that they can answer any questions or accurately incorporate feedback. As this PR does not fit those criteria, I am going to close it now.

binarydesu:
Hi, the Human here.
Thank you for your response, I can totally understand your point.
I just wanted to play some lancer with my friend on foundry without downgrading foundry or running multiple versions at the same time.

I will refrain from further commenting or publishing PRs in this repo, however I would keep the fork up for me or if other people want to temporarily use it for v14, until you find the time to complete the migration.
Also, to prevent this from happening in the future, I would recommend including a disclaimer in the readme. I imagine such incidents will happen more often in the future.

Thank you for developing the system!

PR #973, opened 2026-08-16
Base system localization refactor

annoyedb:
This is **the first part** of the Foundry system localization refactor and a precursor step to supporting [LCP Language Patches](https://github.com/massif-press/lancer-data/wiki/LCP-Language-Patches). Nothing regarding LCP Language Packs has been directly implemented in this PR.

The goal of this update is to clear up technical debt of some hard-coded items as well as 'standardizing' localization where I can (exceptions in the TL;DR)

<details>

<summary>Inside this are my personal notes regarding some of the decisions</summary>

### `lancer.common`
If it can be used ['a couple', 'a few', your judgement tbh] of either: compendium, item, actor, canvas, or document context; then it should be in the `common` context

In hindsight this should have actually been called 'glossary' like how COMP/CON does

### Redundancy
I understand vaguely that languages are a little complicated and to this end, you may notice a few redundancies in the refactor. I tried to compartmentalize these terms as concepts in the context of Lancer rather than their strictly by spelling, since other languages may have different (read: more distinct) words for these things. For example, if I state 'system' in the context of this project, this could mean: a foundry object key 'system', a player mech's 'system', an npc's 'system' feature, or (with some leeway) a pilot's 'systems' HASE stat. If I were to describe these in the localization file, all of these would have their own translation, even if it were the same literal string

### Compounding text
Never compound localization text together. Doing so is a presumption that assumes every language follows the generic structure of the English language and may end up being awkward to read for the affected (plenty of East-Asian languages). The only exception is in the presentation of empiric data, in which case it can be posited that it's 'fine', but it should still be avoided

### Key Tails
All localization keys should terminate in either a minimum `label` key or some other common descriptor to facilitate Foundry's i18n implementation better and maintain organizational consistency. Basically, seeing a terminating key should be immediately recognizable as a dead-end in every localization string.

An exception is ui.notifications, where other tails are basically impossible, and lets the stay in one spot. The only issue is that now you need to be extra verbose in the key names

### Defaults
`short`ened text is capitalized by default. All other text is generally in a format that retains casing information best; I arbitrarily chose titlized. There is no reason for this other than the hope of maintaining consistency, which in turn makes intent clear whenever `upper-case` or such is used in a `.hbs` file. The intent was to "let strings be strings", as I generally dislike correlating stylization and information as one.

### Casing
Casing should be handled all in one way or another. My personal preference is to use CSS' casing styling rules which automatically handles it in the rendering layer, thus any errors won't be fatal and potentially real data does not need to be modified. Implementing this is a bit out of scope however, so I chose to refactor these transformations to be at the `hbs` level, which still achieves the goal of compartmentalizing stylization away from functional code. Again, I just dislike correlating styling and information.

### Technical words
An exception to what I've defined are technical words. Things describing either Foundry or COMP/CON specific key words, such as "LID", "UUID", etc. These are typically hard-coded, but try to note down their use where possible.
</details>

There are probably a few places I broke my own rules, but I tried to maintain consistency where I could.

My take on this isn't going to be perfect, so if there are any concerns I'm obviously open to the feedback. I've also more than likely missed a few spots :grimacing: but it should be overall in a better spot than it was

---

### TL;DR
Functions/handlebars/templates updated:
* actor sheets - all of them
* item sheets - all of them
* chat cards - all of them
* `checkItemLimited` - refactored the `friendly_entry_type_name` now `friendlyEntryTypeName` function
* LCP manager - as a bonus I refactored the LCPManager to use Svelte 5 runes syntax, amongst a few other things
* Pilot importer - I just added some type hardening to the v3 side, getting rid of the `@ts-ignore` I added
* `ui.notifications` - wherever they are

Functions/handlebars/templates avoided:
* macro templates - these look generated or something so I didn't want to bother
* Svelte components other than the LCP manager

Main `en.json` updates:
* normalized _most_ keys to use camelCase
* most keys are now tailed by 'label', 'hint', 'description', etc
* settings localizations is now grouped

`en.json` keys avoided:
* TYPES - [Foundry requirement](https://foundryvtt.com/api/modules/foundry.data.html#localization)
* `LANCERINITIATIVE` - no idea what this is
* `compendium` - no idea how localization is handled in system.json and that file kind of scares me
* `tables` - it's fine as it is
* `squaregriddiagonals` - no idea what this is
* `initiative` - seems like a module string? Kind of weird that it's in the system, but I decided to leave it alone just in case the module relies on the strings being here
* `statusIconsNames` - no idea what this is


annoyedb:
I just realized one of my commit messages says supper :sob::sob: I got hungry

dprint.json
```diff
@@ -18,6 +18,7 @@
     "html.void.selfClosing": false
   },
   "excludes": [
+    "**/*.hbs",
```
Eranziel: Why exclude hbs? One of the main reasons for switching to dprint was because it can keep hbs formatted consistently.

public/templates/actor/deployable.hbs
```diff
@@ -33,67 +33,62 @@
     {{! Active Tab }}
     <div class="tab manage" data-group="primary" data-tab="status">
       <div class="wraprow quadruple">
-        {{{stat-edit-max-card "hp" "mdi mdi-heart-outline" "system.hp.value" "system.hp.max"}}}
-        {{{stat-edit-max-card "heat" "cci cci-heat" "system.heat.value" "system.heat.max"}}}
-        {{{stat-edit-rollable-card "burn" "cci cci-burn" "system.burn"}}}
-        {{{stat-edit-card "overshield" "mdi mdi-shield-star-outline" "system.overshield.value"}}}
+        {{{stat-edit-max-card (upper-case (localize "lancer.common.stat.hp.label")) "mdi mdi-heart-outline" "system.hp.value" "system.hp.max"}}}
```
Eranziel: I suggest we upper case all of these via CSS rather than using the hbs helper. Keeps the templates cleaner.

public/lang/en.json
```diff
@@ -1,30 +1,96 @@
 {
   "TYPES": {
     "Actor": {
-      "deployable": "Deployable",
-      "mech": "Mech",
-      "npc": "NPC",
-      "pilot": "Pilot"
+      "deployable": {
+        "label": "Deployable",
+        "plural": "Deployable"
```
Eranziel: This should be Deployables, surely?

public/lang/en.json
```diff
@@ -37,217 +103,820 @@
...
+          "plural": "Techs",
+          "attack": {
+            "short": "T.ATK",
+            "label": "Tech Attack",
+            "plural": "Tech Attacks"
+          }
+        },
+        "sensor": {
+          "short": "SENS",
+          "label": "Sensor",
```
Eranziel: Suggestion: Sensors

public/lang/en.json
```diff
@@ -37,217 +103,820 @@
...
+          }
+        },
+        "speed": {
+          "short": "SPEED",
+          "label": "Speed"
+        },
+        "repair": {
+          "short": "REPR",
+          "label": "repair",
+          "plural": "repairs"
```
Eranziel: These are uncapitalized for some reason?

